PR #3, opened 2018-10-18
Print to stderr instead of stdout

achew22:
This means that you can pipe output from bazelisk query and it will be parseable.


philwo:
Thanks! :)

I verified that @achew22 signed Google's CLA.

PR #4, opened 2018-10-24
Do not print which bazel version is being used

Globegitter:
This is very noise as it prints that on each invokation and we define the strict version in .bazelversion anyway.

Possibly it would make sense to keep this for the latest->version resolve? But even then if someone is unsure of the version they could just run `bazel version`

### DIFF
--- a/bazelisk.py
+++ b/bazelisk.py
@@ -104,7 +104,6 @@ def main(argv=None):
 
     bazel_version = decide_which_bazel_version_to_use()
     bazel_version = resolve_version_label_to_number(bazelisk_directory, bazel_version)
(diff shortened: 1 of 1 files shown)


Globegitter:
This would also mostly fix #3 as now stdout is only "polluted" on first invocation / version upgrade.

natpen:
I agree with this sentiment, and think printing the version on every invocation is unnecessary noise. Printing new information when doing new things (like downloading a version of bazel) is good, but bazelisk-derived output should otherwise mirror the semantics of bazel itself, to the extent possible.

philwo:
Thanks @Globegitter! I agree. :)

(Note to self: I also verified that the CLA has been signed.)

Issue #2, opened 2018-10-17
Potential for improved excellence on slow network


hlopko:
Dear Philwo,

Thank you for fantastic script! I'm operating on a extremely moody network currently, and at times of slowly flying pigeons bazelisk and intellij fight against each other with this error message:

```
Traceback (most recent call last):
  File "/home/m/bin/bazel", line 116, in <module>
    sys.exit(main())
  File "/home/m/bin/bazel", line 113, in main
    return subprocess.Popen([bazel_path] + argv[1:], close_fds=True).wait()
  File "/usr/lib/python3.6/subprocess.py", line 709, in __init__
    restore_signals, start_new_session)
  File "/usr/lib/python3.6/subprocess.py", line 1344, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
OSError: [Errno 26] Text file busy: '/home/m/.bazelisk/bin/bazel-0.18.0-linux-x86_64'
```

In addition, I'm starring at `Downloading https://releases.bazel.build/0.18.0/release/bazel-0.18.0-linux-x86_64...` for a while now, and I'm wondering if it's doing anything, maybe progress bar would be nice :P

Best regards, and thank you!
big fan,
m.

philwo:
Oh, wow. I didn't think I'd run into my favorite error "text file busy" so quickly :D

I think we can solve this with a lockfile. I'll have a look.

mhlopko:
Hmm apparently it's enough to ctrl-c during the download to bring bazelisk into broken `FATAL: Failed to open Bazel as a zip file: (error: 5): Input/output error` state :)

philwo:
Yeah, I noticed this too, yesterday... it keeps the half-downloaded Bazel file and will try to execute it the next time you run Bazelisk. Should also be easy to fix by downloading into a temporary file first and only moving it over, when done (and then ignoring a potential "file already exists" error that might happen on Windows, I guess). :)

philwo:
This should be fixed - https://github.com/philwo/bazelisk/pull/49 takes care of the corrupted download on Ctrl-C in the Python version. (The Go version already included a fix for this from the beginning.)

I think this also solves the "text file busy" error, but let me know if it happens again!

PR #5, opened 2018-10-24
Add progress bar

Globegitter:
Addresses part of #2 

Example looks:
```
Downloading https://releases.bazel.build/0.17.2/release/bazel-0.17.2-linux-x86_64
[############################################################] 160.5/160.5MB
```


meisterT:
@philwo it would be great to have a progress bar. I was using the go version today and it took three minutes downloading the bazel binary on my connection (who knows why...).

PR #9, opened 2018-11-12
Handle kb interrupts gracefully

natpen:
Before this, if you’re doing something like runinng tests and
then hit ctrl-c, the python dies first, and then the still-running
bazel process continues to write things to stdout until it exits itself.

Note that this bug only affected osx. on linux the behavior upon
ctrl-c was identical to a non-bazelisk invocation of bazel. after this
change, behavior upon ctrl-c is identical to a non-bazelisk
invocation of bazel, for both osx and linux.

### DIFF
--- a/bazelisk.py
+++ b/bazelisk.py
@@ -117,7 +117,11 @@ def main(argv=None):
     os.makedirs(bazel_directory, exist_ok=True)
     bazel_path = download_bazel_into_directory(bazel_version, bazel_directory)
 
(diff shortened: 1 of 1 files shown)


philwo:
Thanks, that's a nice improvement!

I wonder if we should call this in a loop though? With the current fix, the same issue would happen again once the user presses Ctrl-C twice. Bazel's behavior is to gracefully stop the build the first and second time Ctrl-C is pressed and to forcefully shutdown the server as quickly as possible when it's pressed the third time.

PR #10, opened 2018-11-17
Add support for release candidate versions

jin:
This adds support for `rc` versions, like this:

```
$ USE_BAZEL_VERSION=0.20.0rc1 bazelisk build //src:bazel
Using Bazel 0.20.0rc1...
Downloading https://releases.bazel.build/0.20.0/rc1/bazel-0.20.0rc1-darwin-x86_64...
```

### DIFF
--- a/bazelisk.py
+++ b/bazelisk.py
@@ -91,9 +91,25 @@ def determine_bazel_filename(version):
 
     return "bazel-{}-{}-{}".format(version, operating_system, machine)
 
+def determine_release_or_rc(version):
+    parts = version.lower().split("rc")
+    if len(parts) == 1:
+        # e.g. ("0.20.0", "release") for 0.20.0
+        return (version, "release")
+    elif len(parts) == 2:
+        # e.g. ("0.20.0", "rc2") for 0.20.0rc2
+        return (parts[0], "rc" + parts[1])
+    else:
+        raise Exception("Invalid version: {}. Versions must be in the form <x>.<y>.<z>[rc<rc-number>]".format(version))
+
+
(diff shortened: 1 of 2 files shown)


bazelisk.py
```diff
@@ -91,9 +91,25 @@ def determine_bazel_filename(version):
...
+        # e.g. ("0.20.0", "rc2") for 0.20.0rc2
+        return (parts[0], "rc" + parts[1])
+    else:
+        raise Exception("Invalid version: {}. Versions must be in the form <x>.<y>.<z>[rc<rc-number>]".format(version))
+
+
 def download_bazel_into_directory(version, directory):
     bazel_filename = determine_bazel_filename(version)
-    url = "https://releases.bazel.build/{}/release/{}".format(version, bazel_filename)
+    (parsed_version, release_or_rc) = determine_release_or_rc(version)
```
philwo: I think we could replace this with a regular expression, which would also correctly handle exceptions like the recent "0.20rc2" where the patch level version was accidentally missing:

```
(parsed_version, rc) = re.match(r'(\d*\.\d*(?:\.\d*)?)(rc\d)?', rc).groups()
```

Then use it inside a method that computes the URL:

```
def determine_url(version, bazel_filename):
  (parsed_version, rc) = re.match(r'(\d*\.\d*(?:\.\d*)?)(rc\d)?', rc).groups()
  return "https://releases.bazel.build/{}/{}/{}".format(parsed_version,
                                                        rc if rc else "release",
                                                        bazel_filename)
```

philwo:
Actually, let me just merge this and I'll do some reformatting anyway in the next commits. :)

PR #11, opened 2018-11-18
On keyboard interrupt, wait for Bazel to terminate.

philwo:
From @natpen in #9:
"Before this, if you’re doing something like runinng tests and
then hit ctrl-c, the python dies first, and then the still-running
bazel process continues to write things to stdout until it exits itself.""


philwo:
@natpen I think this should solve the problem where Bazelisk terminates early when a user presses Ctrl-C multiple times. What do you think?

natpen:
LGTM! Thanks for the improvement - I didn't even realize Bazel had provisions for handling subsequent `ctrl-c`s differently :stuck_out_tongue_closed_eyes:

PR #14, opened 2018-12-15
Fix bugs on Windows and 64-bit architectures

jbduncan:
This PR fixes https://github.com/philwo/bazelisk/issues/13.

As per the commit message, the changes introduced by this PR do the following:

```
- Correctly fetch Bazel releases ending with ".exe" rather than without
  when downloading on Windows.
- Recognise "amd64" and "AMD64" machine architectures as aliases for
  "x86_64" when downloading Bazel.
```

### DIFF
--- a/bazelisk.py
+++ b/bazelisk.py
@@ -34,8 +34,6 @@ except ImportError:
 
 ONE_HOUR = 1 * 60 * 60
 
-OPERATING_SYSTEM = platform.system().lower()
-
 
 def decide_which_bazel_version_to_use():
   # Check in this order:
@@ -99,14 +97,15 @@ def determine_bazel_filename(version):
     raise Exception('Unsupported machine architecture "{}". '
                     'Bazel currently only supports x86_64.'.format(machine))
 
-  if OPERATING_SYSTEM not in ('linux', 'darwin', 'windows'):
+  operating_system = platform.system().lower()
+  if operating_system not in ('linux', 'darwin', 'windows'):
     raise Exception('Unsupported operating system "{}". '
                     'Bazel currently only supports Linux, macOS and Windows.'
-                    .format(OPERATING_SYSTEM))
+                    .format(operating_system))
 
-  if machine == 'amd64':
-    machine = 'x86_64'
-  return 'bazel-{}-{}-{}'.format(version, OPERATING_SYSTEM, machine)
+  filename_ending = '.exe' if operating_system == 'windows' else ''
+  return 'bazel-{}-{}-{}{}'.format(version, operating_system, machine,
+                                   filename_ending)
 
 
 def normalized_machine_arch_name():
@@ -120,10 +119,8 @@ def determine_url(version, bazel_filename):
   # Split version into base version and optional additional identifier.
   # Example: '0.19.1' -> ('0.19.1', None), '0.20.0rc1' -> ('0.20.0', 'rc1')
   (version, rc) = re.match(r'(\d*\.\d*(?:\.\d*)?)(rc\d)?', version).groups()
-
-  url_ending = '.exe' if OPERATING_SYSTEM == 'windows' else ''
   return "https://releases.bazel.build/{}/{}/{}".format(
-      version, rc if rc else "release", bazel_filename) + url_ending
+      version, rc if rc else "release", bazel_filename)
 
 
 def download_bazel_into_directory(version, directory):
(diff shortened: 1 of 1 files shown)


bazelisk.py
```diff
@@ -34,6 +34,8 @@
 
 ONE_HOUR = 1 * 60 * 60
 
+OPERATING_SYSTEM = platform.system().lower()
```
philwo: I see why you added this, but I think when the second comment is addressed, we can remove this cached global again and just use the local variable as before.

bazelisk.py
```diff
@@ -92,26 +94,36 @@ def resolve_version_label_to_number(bazelisk_directory, version):
...
+    machine = 'x86_64'
+  return machine
 
 
 def determine_url(version, bazel_filename):
   # Split version into base version and optional additional identifier.
   # Example: '0.19.1' -> ('0.19.1', None), '0.20.0rc1' -> ('0.20.0', 'rc1')
   (version, rc) = re.match(r'(\d*\.\d*(?:\.\d*)?)(rc\d)?', version).groups()
+
+  url_ending = '.exe' if OPERATING_SYSTEM == 'windows' else ''
```
philwo: I think it'd be easier to understand the flow of the data if `determine_bazel_filename` simply returned the correct filename with the .exe suffix, instead of adding it here. Otherwise wouldn't we also save it locally to a file not with the .exe suffix and how can we then even execute it on Windows?

bazelisk.py
```diff
@@ -92,26 +94,36 @@ def resolve_version_label_to_number(bazelisk_directory, version):
...
 
-  operating_system = platform.system().lower()
-  if operating_system not in ('linux', 'darwin', 'windows'):
+  if OPERATING_SYSTEM not in ('linux', 'darwin', 'windows'):
     raise Exception('Unsupported operating system "{}". '
                     'Bazel currently only supports Linux, macOS and Windows.'
-                    .format(operating_system))
+                    .format(OPERATING_SYSTEM))
+
+  if machine == 'amd64':
```
philwo: I think we no longer need this, because you pulled out this normalizing logic into the `normalized_machine_arch_name` method?

bazelisk.py
jbduncan: > Otherwise wouldn't we also save it locally to a file not with the .exe suffix and how can we then even execute it on Windows?

Ah, good question. I've just realised that Bazelisk saves Bazel as `~/.bazelisk/bin/bazel-0.20.0-windows-x86_64`, without a `.exe`, and yet it still works (which I assume is because Bazelisk still reads and runs it as if it were a Windows executable).

I agree that it'd be easier to understand if `determine_bazel_filename` returned the correct filename itself rather than delegating it to `determine_url`, so I'll make that change. :+1:

bazelisk.py
jbduncan: Oops, indeed! I'm removing this locally.

bazelisk.py
jbduncan: Indeed! So I'm removing it locally and I will push as soon as I am ready. :)

jbduncan: Done.

bazelisk.py
jbduncan: Done.

bazelisk.py
jbduncan: Done.

jbduncan:
I believe that I've addressed all of your feedback now, @philwo. :)

philwo:
Thanks so much. Looks great! :)

jbduncan:
You're very welcome! It's my pleasure. ^_^

ag559052:
6727905

- [ ] - []()@b6727905b0b8c0fe67fb4de1be3c4816640b372d6727905

# - [x] - []()@jbduncan **_[]()

Introducción a OneDrive.pdf

_**

ag559052 (approved):
6727905

- [ ] - []()@b6727905b0b8c0fe67fb4de1be3c4816640b372dhttps://github.com/bazelbuild/bazelisk/assets/162828811/b816eb45-4cd7-4b73-8b5c-50fbc4b38cb5b6727905b0b8c0fe67fb4de1be3c4816640b372d202404054816640https://blog.google/products/google-play/teacher-approved-apps/.bazelci/config.ymlBienvenido al portapapeles de Gboard; todo texto que copies se # - - [ ] guardará aquí.@ag559052 @ag559052 @jbduncan

PR #24, opened 2019-01-30
bazelisk.go: Correctly handle failed downloads.

philwo:
Progress towards #2.

### DIFF
--- a/bazelisk.go
+++ b/bazelisk.go
@@ -208,26 +208,46 @@ func downloadBazel(version, directory string) string {
 	destinationPath := path.Join(directory, filename)
 
 	if _, err := os.Stat(destinationPath); err != nil {
-		log.Printf("Downloading %s...", url)
-		f, err := os.Create(destinationPath)
+		tmpfile, err := ioutil.TempFile(directory, "download")
 		if err != nil {
-			log.Fatalf("Could not create %s: %v", destinationPath, err)
+			log.Fatalf("Could not create temporary file: %v", err)
 		}
-		defer f.Close()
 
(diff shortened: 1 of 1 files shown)


bazelisk.go
```diff
@@ -208,26 +208,46 @@ func downloadBazel(version, directory string) string {
...
 
 	if _, err := os.Stat(destinationPath); err != nil {
-		log.Printf("Downloading %s...", url)
-		f, err := os.Create(destinationPath)
+		tmpfile, err := ioutil.TempFile(directory, "download")
 		if err != nil {
-			log.Fatalf("Could not create %s: %v", destinationPath, err)
+			log.Fatalf("Could not create temporary file: %v", err)
 		}
-		defer f.Close()
```
fweikert: What about adding

defer func() {
		err := tmpfile.Close()
		if err != nil {
			os.Remove(tmpfile.Name())
		}
	}()

and removing the duplicate code below? I guess you only need to keep the close before renaming it.

philwo: Now *that's* a nice idea. I didn't even know you could pass a "func" to defer. Cool, thanks!

philwo: One tricky thing is that `log.Fatalf` calls `os.Exit`, which will immediately cause the program to exit without running deferred functions, so the temporary file will not be deleted in error cases.

I wonder if it's better to use `panic` here? 🤔 But that prints a stack trace, which might be irrelevant for the user.

OK, maybe returning an error with additional information that wraps the original error is actually better. I'll try that.

philwo: Done!

PR #36, opened 2019-02-27
Fix error code and argument order

laurentlb:
- Exit with error code 1 if the build fails
- Fix a bug in a print message
- Insert arguments before "--" if any (otherwise, put them at the end)

Fixes https://github.com/philwo/bazelisk/issues/34

### DIFF
--- a/bazelisk.go
+++ b/bazelisk.go
@@ -415,7 +415,7 @@ func migrate(bazelPath string, baseArgs []string, newArgs []string) {
 	}
 	if exitCode != 0 {
 		fmt.Printf("Failure: Command failed, even without incompatible flags.\n")
-		os.Exit(1)
+		os.Exit(exitCode)
 	}
 
 	// 3. Try with each flag separately.
--- a/bazelisk.go
+++ b/bazelisk.go
@@ -370,10 +370,30 @@ func getIncompatibleFlags(bazeliskHome, resolvedBazelVersion string) ([]string, 
 	return result, nil
 }
 
+// insertArgs will insert newArgs in baseArgs. If baseArgs contains the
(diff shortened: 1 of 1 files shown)


laurentlb:
cc @meteorcloudy

laurentlb:
I don't know what's failing.
Tests pass locally.

philwo:
The tests are failing, because the `releases_for_tests.json` somehow isn't used, thus Bazelisk now picks up a newer Bazel release that was released in the meantime, thus the hardcoded golden version strings in the test no longer match.

I don't know why, though :/

bazelisk.go
```diff
@@ -387,22 +407,22 @@ func migrate(bazelPath string, baseArgs []string, newArgs []string) {
...
+	fmt.Printf("\n\n--- Running Bazel with no incompatible flags\n\n")
 	fmt.Printf("bazel %s\n", strings.Join(args, " "))
 	exitCode, err = runBazel(bazelPath, args)
 	if err != nil {
 		log.Fatalf("could not run Bazel: %v", err)
 	}
 	if exitCode != 0 {
 		fmt.Printf("Failure: Command failed, even without incompatible flags.\n")
-		os.Exit(0)
+		os.Exit(1)
```
philwo: Can we do `os.Exit(exitCode)` here to preserve Bazel's exit code?

meteorcloudy:
The Bazelisk + Incompatible flags pipeline is basically ready
https://buildkite.com/bazel/bazelisk-plus-incompatible-flags/builds/17

But the result are not entirely correct due to the issues this PR is going to fix.

bazelisk.go
laurentlb: Good point, done.

PR #38, opened 2019-02-28
The bazel binaries are now put inside the user cache dir

mpassell:
Just a quick README fix that I happened to notice.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,4 +81,4 @@ variable](https://github.com/golang/go/wiki/SettingGOPATH).
 ## FAQ
 
 ### Where does Bazelisk store the downloaded versions of Bazel?
-It creates a directory called ".bazelisk" inside your home directory and will store them there. Feel free to delete this directory at any time, as it can be regenerated automatically when required.
+It creates a directory called "bazelisk" inside your [user cache directory](https://golang.org/pkg/os/#UserCacheDir) and will store them there. Feel free to delete this directory at any time, as it can be regenerated automatically when required.
--- a/bazelisk.py
+++ b/bazelisk.py
@@ -169,6 +169,16 @@ def resolve_latest_version(version_history, offset):
     return version_history[offset]
 
 
+def get_operating_system():
+    operating_system = platform.system().lower()
+    if operating_system not in ("linux", "darwin", "windows"):
+        raise Exception(
+            'Unsupported operating system "{}". '
+            "Bazel currently only supports Linux, macOS and Windows.".format(operating_system)
+        )
+    return operating_system
+
+
 def determine_bazel_filename(version):
     machine = normalized_machine_arch_name()
     if machine != "x86_64":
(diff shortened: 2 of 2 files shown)


jbduncan:
Wait, isn't this a bug? I ask because last time I used the Python version of Bazelisk, it stored the Bazel binaries in `.bazelisk`, so I'm a bit surprised that the Go version stores them in a different location. Should we not make the Python and Go versions consistent instead? 🤔

mpassell:
You're right - https://github.com/philwo/bazelisk/blob/master/bazelisk.py#L256 . I've recently heard people advocating for the use of the user cache dir for cases like this, partly since it allows people to choose where the files go (at least on Linux) by setting `XDG_CACHE_HOME`.

jbduncan:
Sounds good to me, although I do believe that an explicit goal of the Python version is to avoid using third-party modules so that the script can be used in a self-contained fashion. I could be wrong about this though. :)

@philwo Do you have any thoughts regarding all this?

philwo:
I would prefer to change the Python version to use the same approach as Go:

> On Unix systems, it returns $XDG_CACHE_HOME as specified by https://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html if non-empty, else $HOME/.cache. On Darwin, it returns $HOME/Library/Caches. On Windows, it returns %LocalAppData%. On Plan 9, it returns $home/lib/cache.

I think this is preferable to using the home directory of the user directly and also seems to give nice results on all supported platforms.

Let's not add a third-party dependency for this, though, due to the reason that @jbduncan mentioned. If someone wants to give this a try, I'm happy to review a PR. :)

philwo:
@mpassell Thanks for the PR! I'm happy to merge this, but considering that the Python behavior still has a different behavior, should we either mention both, or wait until we implemented the same logic in the Python version?

mpassell:
You're welcome. :) If the logic change in Python is as simple as I'm guessing, holding off on this until they're consistent makes sense. Oh, and fair enough about the module. It probably doesn't do much more than reading in a few env vars, anyway. I can try to take a crack at this tomorrow, but the result may not be very idiomatic, since I'm no Python expert.

mpassell:
I'm getting back to this now. I guess I'll just port https://github.com/golang/go/blob/4b05dc91b0fe99724721e3545807f3b8f1310c52/src/os/file.go#L347 to the Python version for consistency. Does bazelisk support Windows or Plan9? (I'm guessing maybe on the former, unlikely on the latter. 😄 )

philwo:
Thanks! Yes, Windows is supported, Plan9 not (yet?) 😄

The BAZELISK_HOME should still be supported - it's used in the interesting tests and maybe even on Bazel's CI to override the location.

mpassell:
Let me know what you think of the code I pushed up. I'm actually going to repush it since it's using the wrong git identity for me, but the code won't change.

philwo:
Thanks so much @mpassell! Looks great. :)

For the record, I verified that the Google CLA has been signed, so I can merge this PR.

mpassell:
Great! Yup, I signed the CLA so that I could make a little doc fix to Skylib. So far, that PR hasn't ended up in my writing any code. 😆

PR #41, opened 2019-03-08
Get up to 100 flags (instead of 30).

laurentlb:
Instead of hard-coding the limit, we could also make multiple requests,
one for each page (https://developer.github.com/v3/#pagination).

But we don't expect the number of flags to grow indefinitely, so a limit
should work.

Fixes https://github.com/philwo/bazelisk/issues/40


laurentlb:
cc @meteorcloudy

PR #46, opened 2019-04-03
execute tools/bazel if it exists

### LINKED ISSUE #1: Add support for Autogazelle
See https://github.com/bazelbuild/bazel-gazelle/tree/master/cmd/autogazelle

hperl:
This change enables the combination of bazelisk and wrappers like autogazelle.

Fixes #1

### DIFF
--- a/bazelisk.py
+++ b/bazelisk.py
@@ -44,6 +44,8 @@ BAZEL_GCS_PATH_PATTERN = (
 
 SUPPORTED_PLATFORMS = {"linux": "ubuntu1404", "windows": "windows", "darwin": "macos"}
 
+TOOLS_BAZEL_PATH = "./tools/bazel"
+
 
 def decide_which_bazel_version_to_use():
     # Check in this order:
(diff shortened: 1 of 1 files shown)


fweikert:
No objections once the test passes :)

bazelisk.py
```diff
@@ -273,8 +273,13 @@ def maybe_makedirs(path):
 
 
 def execute_bazel(bazel_path, argv):
+    env = {}
+    # try to execute bazel wrapper in ./tools first.
+    if os.path.exists("./tools/bazel"):
+        env["BAZEL_REAL"] = bazel_path
+        bazel_path = "./tools/bazel"
```
fweikert: Nit: maybe move ./tools/bazel into a constant

bazelisk.py
```diff
@@ -273,6 +275,10 @@ def maybe_makedirs(path):
 
 
 def execute_bazel(bazel_path, argv):
+    # try to execute bazel wrapper in ./tools first.
+    if os.path.exists(TOOLS_BAZEL_PATH):
```
philwo: I don't think we can assume that our current working directory is in the workspace root (it's allowed to run Bazel from any subdirectory of a workspace), so we should instead use find_workspace_root() to get the root directory and then use `os.path.join` to get to the path to $root/tools/bazel.

philwo:
Thanks for the contribution!

I merged https://github.com/philwo/bazelisk/pull/53 instead, which addressed the open comments from this one and also included a Go version - hope you don't mind.

PR #49, opened 2019-04-17
Use tempfile for bazel download in bazelisk.py

robbertvanginkel:
To address https://github.com/philwo/bazelisk/issues/43, download bazelisk into a tempfile and only move it to the cache directory after the download fully succeeds.

Open to suggestions for how to test this.

### DIFF
--- a/bazelisk.py
+++ b/bazelisk.py
@@ -24,6 +24,7 @@ import re
 import shutil
 import subprocess
 import sys
+import tempfile
 import time
 
 try:
@@ -226,9 +227,12 @@ def download_bazel_into_directory(version, is_commit, directory):
     destination_path = os.path.join(directory, bazel_filename)
     if not os.path.exists(destination_path):
         sys.stderr.write("Downloading {}...\n".format(url))
-        with closing(urlopen(url)) as response:
-            with open(destination_path, "wb") as out_file:
-                shutil.copyfileobj(response, out_file)
+        with tempfile.NamedTemporaryFile(prefix="bazelisk", dir=directory, delete=False) as t:
+            with closing(urlopen(url)) as response:
+                shutil.copyfileobj(response, t)
+            t.flush()
+            os.fsync(t.fileno())
+        os.rename(t.name, destination_path)
     os.chmod(destination_path, 0o755)
     return destination_path
 


bazelisk.py
```diff
@@ -226,9 +227,12 @@ def download_bazel_into_directory(version, is_commit, directory):
...
     if not os.path.exists(destination_path):
         sys.stderr.write("Downloading {}...\n".format(url))
-        with closing(urlopen(url)) as response:
-            with open(destination_path, "wb") as out_file:
-                shutil.copyfileobj(response, out_file)
+        with tempfile.NamedTemporaryFile(prefix="bazelisk", delete=False) as t:
+            with closing(urlopen(url)) as response:
+                shutil.copyfileobj(response, t)
+            t.flush()
+            os.fsync(t.fileno())
```
philwo: Why is the flush and fsync needed? Closing the handle should be enough, shouldn't it?

philwo:
Thank you, this looks good! :)

Would you mind signing the Google CLA, so that I can accept your contribution? Sorry to bother you with this, but otherwise I won't be able to move this project into the Bazel organization in the future. It's here: https://cla.developers.google.com/

bazelisk.py
robbertvanginkel: As I understand it, `fsync` is needed before a rename to guarantee that `os.rename` doesn't create a 0 length file on posix systems. I'm going off the discussion in https://github.com/golang/go/issues/22397#issuecomment-380831736.

Updated this to use `os.rename` instead of `shutil.move` as a move across filesystems would become a copy and not be atomic. To make sure the tempfile is on the same filesystem, changed to create the temp file in the bazelisk cache directory.

robbertvanginkel:
Updated the commit to a different email, on which I'd already signed the cla. Let me know if that works.

akshayjshah:
We'll run into the same issue in the Go version, right? I think we should be able to use Michael Stapelberg's `github.com/google/renameio` to do the same thing there (but with a little less pain).

robbertvanginkel:
@philwo can you take another look?

bazelisk.py
philwo: Ah! That's a good point. :) Thanks!

PR #50, opened 2019-04-18
Remove Homebrew TODO

HackAttack:
This has been done (bazelbuild/homebrew-tap#40).


philwo:
Thank you!

Note to self: I verified that @HackAttack signed the Google CLA.

PR #56, opened 2019-05-16
Fixed searching of workspace on Windows.

excitoon:
On Windows Bazelisk previously only read `.bazelversion` if you are in the same directory.
@philwo


excitoon:
On my platform (go version go1.12 windows/amd64), `path.Dir` only takes into account simple slashes, and `filepath.Dir` uses OS path separator.

Issue #18, opened 2019-01-02
Support GitHub API Token


dflems:
If you're on a shared network, you can sometimes see messages like this from the API:

```
$ curl https://api.github.com/repos/bazelbuild/bazel/releases
{
  "message": "API rate limit exceeded for 123.45.67.89. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)",
  "documentation_url": "https://developer.github.com/v3/#rate-limiting"
}
```

Homebrew (for instance) has support for supplying a token via `HOMEBREW_GITHUB_API_TOKEN`, making it more likely that the API request won't get rate limited. It would be nice for `bazelisk` to support this as well, perhaps via a similar convention (`BAZELISK_GITHUB_API_TOKEN`) to perform authenticated requests instead.

dflems:
Ah I just noticed bazelisk _only_ pulls the releases list from `github.com` and nothing else. Maybe it would be easier to publish something like `https://releases.bazel.build/latest-version.txt` on your end and pull that down instead :)

keith:
@philwo any ideas on a solution for this? I could dedicate some time to this if we had a clear path

philwo:
Both ideas sound good - using a GitHub token (if available) and storing that information on our own host. I can handle the latter part, but integrating it into our release pipeline will take some time.

@keith do you want to look into using a GitHub token? Sounds like a good and quick fix for this. I'm not sure where people would usually store these tokens - environment variable? If that's the usual way, sure, why not. :)

keith:
I've submitted an attempt at this here https://github.com/bazelbuild/bazelisk/pull/62

PR #62, opened 2019-05-29
Add support for GitHub auth token

keith:
This adds support for BAZELISK_GITHUB_TOKEN which is a GitHub token used
for API requests when it is set. This fixes the issue of being on a
shared network and getting rate limited when using bazelisk.

Fixes https://github.com/bazelbuild/bazelisk/issues/18

### DIFF
--- a/bazelisk.go
+++ b/bazelisk.go
@@ -105,8 +105,18 @@ type release struct {
 	Prerelease bool   `json:"prerelease"`
 }
 
-func readRemoteFile(url string) ([]byte, error) {
-	res, err := http.Get(url)
+func readRemoteFile(url string, token string) ([]byte, error) {
+	client := &http.Client{}
+	req, err := http.NewRequest("GET", url, nil)
+	if err != nil {
+		return nil, fmt.Errorf("could not create request: %v", err)
+	}
+
+	if token != "" {
+		req.Header.Set("Authorization", "token "+token)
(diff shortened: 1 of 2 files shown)


bazelisk.go
```diff
@@ -105,8 +105,18 @@ type release struct {
...
-	res, err := http.Get(url)
+func readRemoteFile(url string, token string) ([]byte, error) {
+	client := &http.Client{}
+	req, err := http.NewRequest("GET", url, nil)
+	if err != nil {
+		return nil, fmt.Errorf("could not create request: %v", err)
+	}
+
+	if token != "" {
+		req.Header.Set("Authorization", "token "+token)
```
keith: This logic might be a bit too GitHub specific, if so I'm open to trying to break it up

bazelisk.go
```diff
@@ -140,7 +150,7 @@ func maybeDownload(bazeliskHome, url, filename, description string) ([]byte, err
 	}
 
 	// We could also use go-github here, but I can't get it to build with Bazel's rules_go and it pulls in a lot of dependencies.
-	body, err := readRemoteFile(url)
```
keith: By doing this I've taken over `maybeDownload`, I'm happy to rename it if that makes sense.

keith:
Sounds great, at that point we could delete this

PR #66, opened 2019-06-02
Print bazelisk version formation when running 'bazel version'

### LINKED ISSUE #65: Version information
One feature that I'm missing, is a way to troubleshoot if a user is using `bazelisk` vs the raw `bazel`.

This is especially useful if the `bazelisk` binary is installed as `bazel`.

My proposal is that `bazelisk` should intercept `bazelisk version` and prepend or append the bazelisk version to the output, like so:

```
$ bazelisk version
Bazelisk version: 0.0.5
Build label: 0.26.0
Build target: bazel-out/darwin-opt/bin/src/main/java/com/google/devtools/build/lib/bazel/BazelServer_deploy.jar
Build time: Tue May 28 08:38:24 2019 (1559032704)
Build timestamp: 1559032704
Build timestamp as int: 1559032704
```

zegl:
The version information is detected from git tags, with a fallback
to the latest git commit.

Fixes #65

### DIFF
--- a/bazelisk.go
+++ b/bazelisk.go
@@ -641,7 +641,21 @@ func main() {
 	// print bazelisk version information if "version" is the first argument
 	// bazel version is executed after this command
 	if len(args) > 0 && args[0] == "version" {
-		fmt.Printf("Bazelisk version: %s\n", BazeliskVersion)
+		// Check if the --gnu_format flag is set, if that is the case,
+		// the version is printed differently
+		var gnuFormat bool
+		for _, arg := range args {
+			if arg == "--gnu_format" {
+				gnuFormat = true
+				break
+			}
+		}
+
+		if gnuFormat {
+			fmt.Printf("Bazelisk %s\n", BazeliskVersion)
+		} else {
+			fmt.Printf("Bazelisk version: %s\n", BazeliskVersion)
+		}
 	}
 
 	exitCode, err := runBazel(bazelPath, args)
--- a/bazelisk.go
+++ b/bazelisk.go
@@ -42,6 +42,10 @@ const (
 	wrapperPath = "./tools/bazel"
 )
 
+var (
+	BazeliskVersion = "development"
+)
+
 func findWorkspaceRoot(root string) string {
 	if _, err := os.Stat(filepath.Join(root, "WORKSPACE")); err == nil {
 		return root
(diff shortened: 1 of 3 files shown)


bazelisk.go
```diff
@@ -634,6 +638,12 @@ func main() {
 		}
 	}
 
+	// print bazelisk version information if "version" is the first argument
+	// bazel version is executed after this command
+	if len(args) > 0 && args[0] == "version" {
+		fmt.Printf("Bazelisk version: %s\n", BazeliskVersion)
```
philwo: We should handle the case where `--gnu_format` is passed to `bazel version`. In that case, we have to print a slightly different version syntax: https://www.gnu.org/prep/standards/html_node/_002d_002dversion.html

Maybe like this:
```
$ bazel version --gnu_format
Bazelisk 0.0.5
Bazel 0.26.0
```

stamp.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+# Use the first tag that points to the current HEAD
+# if no tag is found, the latest git commit is used as a fallback
+
+CURRENT_TAG=$(git tag -l --points-at HEAD | head -n1);
```
philwo: Can we remove the trailing semicolons here?

.bazelrc
```diff
@@ -0,0 +1 @@
+build --workspace_status_command $PWD/stamp.sh
```
philwo: I think this should work without `$PWD/`, because Bazel runs the command specified here using bash in the root directory of the workspace. Maybe this also fixes the failure on macOS.

zegl: The tests are only failing on [Windows](https://buildkite.com/bazel/bazelisk/builds/134#29116b83-4a05-4ec3-bc8b-d0962745ffd7).

Do you know of any good ways to create cross-platform `--workspace_status_command` scripts? I've seen that a few projects are using Python, but that is very overkill for what I'd like to do here.

Maybe we should just skip the `.bazelrc`? Adding the workspace data is only really needed when building release artefacts, which I'm assuming is done manually today.

bazelisk.go
zegl: Thanks, I didn't know about the `--gnu_format` argument, I'll implement it. 👍

.bazelrc
philwo: Ah, true :D I looked in the wrong spot.

I'm also not aware of a better mechanism that works cross-platform :neutral_face: 

We can drop the `.bazelrc` file and I'll manually specify that flag when building the release artifacts then. Thanks!

stamp.sh
zegl: Done!

.bazelrc
zegl: I dropped it for now. :)

philwo:
@zegl Thanks a lot! :) I'll merge this!

@fweikert FYI, we should try to change our build.sh script to use Bazel to build the release binaries and verify that the version number has been correctly embedded when we do the next release :)

PR #67, opened 2019-06-04
Fix #61: Do not execute tools/bazel when BAZELISK_SKIP_WRAPPER is set.

### LINKED ISSUE #61: If //tools/bazel points to bazelisk, infinite fork happens
We have been using //tools/bazel to execute bazelisk and then that picks up which version to run via .bazelversion.

Change #53 broke this. 

Would it be possible to add a cmd line flag to disable this feature?

philwo:
Bazelisk will set this environment variable when executing Bazel (or
a wrapper found in <workspace_root>/tools/bazel). This prevents Bazelisk
from executing itself in an infinite loop, in case the wrapper is
Bazelisk or somehow executes Bazelisk.

If you're using a wrapper in <workspace_root>/tools/bazel that downloads
Bazelisk and delegates to it, please set the BAZELISK_SKIP_WRAPPER
environment variable to ensure that Bazelisk directly executes Bazel.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,13 @@ show you which flags can safely enabled, and which flags require a migration.
 You can set `BAZELISK_GITHUB_TOKEN` to set a GitHub access token to use for API
 requests to avoid rate limiting when on shared networks.
 
+If `tools/bazel` exists in your workspace root and is executable, Bazelisk will run this file,
+instead of the Bazel version it downloaded. It will set the environment variable `BAZEL_REAL` to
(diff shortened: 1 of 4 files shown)


philwo:
FYI @kellycampbell - when this is in and you modify your wrapper script to set the BAZELISK_SKIP_WRAPPER environment variable (any value except the empty string is fine), this should solve your problem. :)

I'll do a Bazelisk release as soon as this is submitted.

Issue #73, opened 2019-06-25
Migrate not strict enough for some flags


keith:
In our testing with bazelisk migrate we've noticed that bazelisk doesn't clean up enough between builds to make sure it's correct about whether or not a flag needs to be migrated.

In our specific case testing migrate with 0.27.0 we get an initial failure with all the flags:

```
bazel build --config=QA --apple_platform_type=ios //Modules/... --incompatible_disable_legacy_proto_provider --incompatible_disable_native_android_rules --incompatible_disable_proto_source_root --incompatible_disallow_legacy_java_provider --incompatible_disallow_legacy_javainfo --incompatible_enable_cc_toolchain_resolution --incompatible_make_thinlto_command_lines_standalone --incompatible_no_target_output_group --incompatible_remap_main_repo --incompatible_remove_legacy_whole_archive --incompatible_require_java_toolchain_header_compiler_direct --incompatible_restrict_named_params --incompatible_use_aapt2_by_default --incompatible_windows_native_test_wrapper

Loading:
Loading: 0 packages loaded
Analyzing: 2143 targets (470 packages loaded, 0 targets configured)
Analyzing: 2143 targets (715 packages loaded, 15147 targets configured)
Analyzing: 2143 targets (802 packages loaded, 21675 targets configured)
INFO: Analyzed 2143 targets (802 packages loaded, 24113 targets configured).
Building: checking cached actions
INFO: Found 2143 targets...
INFO: Deleting stale sandbox base /private/var/tmp/_bazel_iosci/4da19724c573aed48a58c78a6f864623/sandbox
[8 / 34] [Prepa] Writing file Modules/MapToastListeners/MapToastListeners.output_file_map.json
[497 / 557] Compiling Swift module GlobalSupportService; 1s local ... (12 actions, 11 running)
[579 / 638] AssetCatalogCompile Modules/CoreUI/CoreUIResources-intermediates/xcassets; 2s local ... (12 actions, 10 running)
[690 / 747] AssetCatalogCompile Modules/CoreUI/CoreUIResources-intermediates/xcassets; 4s local ... (12 actions, 11 running)
[832 / 931] AssetCatalogCompile Modules/CoreUI/CoreUIResources-intermediates/xcassets; 6s local ... (12 actions, 11 running)
[930 / 1,083] AssetCatalogCompile Modules/CoreUI/CoreUIResources-intermediates/xcassets; 9s local ... (12 actions, 11 running)
[...]

PR #75, opened 2019-07-03
Run clean between incompatible flag checks

keith:
Fixes https://github.com/bazelbuild/bazelisk/issues/73

### DIFF
--- a/bazelisk.go
+++ b/bazelisk.go
@@ -540,8 +540,8 @@ func insertArgs(baseArgs []string, newArgs []string) []string {
 }
 
 func cleanIfNeeded(bazelPath string) {
-	bazeliskNoClean := os.Getenv("BAZELISK_CLEAN")
-	if len(bazeliskNoClean) == 0 {
+	bazeliskClean := os.Getenv("BAZELISK_CLEAN")
+	if len(bazeliskClean) == 0 {
 		return
 	}
 
--- a/bazelisk.go
+++ b/bazelisk.go
@@ -539,9 +539,9 @@ func insertArgs(baseArgs []string, newArgs []string) []string {
 	return result
 }
 
-func clean(bazelPath string) {
-	bazeliskNoClean := os.Getenv("BAZELISK_NO_CLEAN")
-	if len(bazeliskNoClean) != 0 {
+func cleanIfNeeded(bazelPath string) {
+	bazeliskNoClean := os.Getenv("BAZELISK_CLEAN")
+	if len(bazeliskNoClean) == 0 {
 		return
 	}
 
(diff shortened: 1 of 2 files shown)


bazelisk.go
```diff
@@ -539,12 +539,29 @@ func insertArgs(baseArgs []string, newArgs []string) []string {
 	return result
 }
 
+func clean(bazelPath string) {
+	bazeliskNoClean := os.Getenv("BAZELISK_NO_CLEAN")
```
keith: We might want to flip this instead and have users opt in since it could be a big time difference

meteorcloudy: Agree.

philwo: The question is: While running migrate without cleaning is obviously faster, does it give any usable results? In theory it should, in practice I’m not at all convinced. 

Incompatible flag flips that don’t change action cache keys will not cause any actions to rerun, even though there might be differences, e.g. in how execution strategy selection works, how execution strategies do their stuff, memory consumption...

What’s our history of detecting and missing breakages with the current behavior?

meteorcloudy: The current behavior actually gives us pretty accurate result in the [incompatible change pipeline](https://buildkite.com/bazel/bazelisk-plus-incompatible-flags/builds/162). I had the same concern when I switched incompatible flags pipeline implementation from running every flag with every project in an individual job to running Bazelisk for all flags in one job. But ever since then, I saw only two times Bazelisk gave wrong result because of not doing bazel clean. The later one actually helped us discover a [bug](https://github.com/bazelbuild/bazel/issues/8693).

laurentlb: If we get incorrect results without clean, that's a bug in Bazel. I'd prefer that we file bugs in that case.

philwo: > If we get incorrect results without clean, that's a bug in Bazel. I'd prefer that we file bugs in that case.

Yes, I agree with that. I believe that all changes to execution strategies will not be adequately tested without doing a "clean". I don't know about other areas, maybe things look better there.

But if you think it's fine to make this opt-in (= tell the user they have to make the decision), then under which circumstances should users opt-in and under which would it be fine to not do so?

Unless we can clearly answer this, I think it's wrong to make this configurable - either doing it without "clean" is OK, then we do it without clean, or it's not OK, then we always do it with "clean" (even though it's slower). How should a user make the correct decision here if not even we can?

laurentlb: I've filed https://github.com/bazelbuild/bazel/issues/8787 to track the specific issue of incompatible_enable_cc_toolchain_resolution.
I've sent an email (https://groups.google.com/forum/#!topic/bazel-dev/UJubXrYxHVM) to discuss the root cause.

hlopko: +1 to what Laurent said, let's fix the root cause and keep Bazelisk assuming Bazel is correct.

keith: Based on this conversation should we keep the default the same and still add this, or not add this at all until we have a better understanding of this specific issue?

philwo: Thanks for the productive discussion everyone :)

@keith Let’s add the flipped version!

keith: Flipped!

philwo:
Code looks good, thanks!

Can you add a quick note to the README about this new feature?

keith:
Added to the readme!

Issue #79, opened 2019-07-15
Support for Installing Custom Bazel Versions 


borkaehw:
We use a fork for Bazel at Lucid Software because we utilize features in Bazel that are still in code review or are too premature for a pull request to the main Bazel repo. 

We would like to be able to use Bazelisk with our fork of the Bazel repo, so we can manage Bazel upgrades while on a custom version of Bazel.

Currently Bazelisk only supports downloading official releases.
```
fmt.Sprintf("https://releases.bazel.build/%s/%s/%s", version, kind, filename)
```

It would be nice to have an option to choose between the normal GitHub release URL like

```
fmt.Sprintf("https://github.com/%s/bazel/releases/download/%s/%s", remote, version, filename)
```

We already have a prototype. We are happy to draft a PR or draft a design document. Thanks for your consideration.

philwo:
Hi @borkaehw,

this is a feature I wanted to add anyway - feel free to send a PR (with accompanying documentation change for the README on how to use it) and we can get it in. :)

Thanks!

borkaehw:
Thanks for your support. Currently I am thinking of something as simple as a `--remote` flag. If the flag is provided like `--remote=the-fork`, it would use the URL `https://github.com/the-fork/bazel/releases/download/...`, otherwise use the original one. It still restricts the use cases to official releases and GitHub releases, but I suppose it would cover most of the cases.

What do you think?

PR #80, opened 2019-07-16
Support for installing custom Bazel version

borkaehw:
Fix: https://github.com/bazelbuild/bazelisk/issues/79

Since I am using `flag`, I believe it will work well for `--strict` and `--migrate`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ In the future we will add support for building Bazel from source at a given comm
 
 ## Other features
 
-The Go version of Bazelisk offers two new flags.
+The Go version of Bazelisk offers three new flags.
 
 `--strict` expands to the set of incompatible flags which may be enabled for the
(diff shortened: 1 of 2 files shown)


borkaehw:
Hey @philwo, I think this is ready for review. I'm pinging you since you showed interest in the issue thread.

borkaehw:
Sorry, I just realized we can't use `flag` since `bazelisk` doesn't know which flag should go to `bazelisk` and which should go to `bazel`. I will close this PR and reopen it with a fix. Thanks.

philwo:
@borkaehw Sorry for the delay! Your PR looked good in general, but the flag parsing was the one thing I wanted to point out.

Maybe we could just use an environment variable? There's precedent for that in Bazelisk, like the USE_BAZEL_VERSION environment variable.

PR #81, opened 2019-07-25
Support for installing custom Bazel version

borkaehw:
Fix: #79

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,12 +15,20 @@ Some ideas how to use it:
 - Check it into your repository and recommend users to build your software via `./bazelisk.py build //my:software`. That way, even someone who has never used Bazel or doesn't have it installed can build your software.
 - As a company using Bazel or as a project owner, add a `.bazelversion` file to your repository. This will tell Bazelisk to use the exact version specified in the file when running in your workspace. The fact that it's versioned inside your repository will then allow for atomic upgrades of Bazel including all necessary changes. If you install Bazelisk as `bazel` on your CI machines, too, you can even test Bazel upgrades via a normal presubmit / pull request. It will also ensure that users will not try to build your project with an incompatible version of Bazel, which is often a cause for frustration and failing builds.
 
-## How does Bazelisk know which version to run?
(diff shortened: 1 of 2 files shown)


borkaehw:
@philwo I believe this works better than the previous one. Please take a look when you get a chance. Thanks.

borkaehw:
@philwo friendly pin here. Any feedback will be appreciated.

README.md
```diff
@@ -34,6 +34,17 @@ Bazelisk currently understands the following formats for version labels:
 
 In the future we will add support for building Bazel from source at a given commit.
 
+## How does Bazelisk download release from a fork?
+
+It uses a simple algorithm:
+- If the environment variable `USE_BAZEL_REMOTE` is set, it will use the fork name specified in the value.
```
philwo: Can we call this `USE_BAZEL_FORK` (and respectively below replace "remote" with "fork" in other places, too)?

While I understand where the name "remote" comes from, it's already a pretty often used term in Bazel (e.g. for "remote execution" and "remote caching"). I worry a bit that there might be confusion around this.

README.md
```diff
@@ -34,6 +34,17 @@ Bazelisk currently understands the following formats for version labels:
...
+- Otherwise, if a `.bazelremote` file exists in the current directory or recursively any parent directory, it will read the file and use the the fork name specified in it.
```
philwo: Instead of using a separate file, what do you think of adding a "fork/" prefix to the .bazelversion file (and general version label parsing) instead? So users could do this:

echo "philwo/latest" > ~/.bazelversion
or
USE_BAZEL_VERSION="philwo/0.28.1"

It would be backwards compatible, too.

README.md
borkaehw: It's a good idea. Thanks.

README.md
borkaehw: This is a good idea too.

borkaehw:
@philwo I have made the changes. The new way looks cleaner. Please take another look.

borkaehw:
Thanks for improving the documentation.

PR #82, opened 2019-07-25
Fix empty .bazelversion download error

borkaehw:
Fix the download error if there is an empty `.bazelversion`.

It would be useful and more robust when users temporarily need the latest version of Bazel, but they don't need to remove `.bazelversion`.


philwo:
Thanks! Yes, seems like a good idea to fallback to the latest release when the string is empty. :)

Issue #89, opened 2019-09-10
Bazelisk does not honor locally installed bazel version


davido:
On our CI docker we install Bazel version from the official Debian image,
that corresponds to the `.bazelversion` file, by using official Bazel version:

https://releases.bazel.build/1.0.0/rc2/bazel_1.0.0rc2-linux-x86_64.deb

In the same time we install bazelisk, version 1.0 on CI docker image.

If I run `bazelisk build :release` and .bazelversion is 1.0.0rc2, and locally the right version of Bazel is installed, I would expect that Bazelisk is smart enough, to use that version.

What happens, is yet another instance of Bazel is downloaded by Bazelisk:

```
davido@wizball:~/projects/gerrit$ bazel version
Build label: 1.0.0rc2
Build target: bazel-out/k8-opt/bin/src/main/java/com/google/devtools/build/lib/bazel/BazelServer_deploy.jar
Build time: Fri Sep 6 12:17:09 2019 (1567772229)
Build timestamp: 1567772229
Build timestamp as int: 1567772229

davido@wizball:~/projects/gerrit$ which bazel     
/usr/bin/bazel

davido@wizball:~/projects/gerrit$ cat .bazelversion 
1.0.0rc2

davido@wizball:~/projects/gerrit$ bazelisk build java/com/google/gerrit/server:server
2019/09/10 21:43:16 Downloading https://releases.bazel.build/1.0.0/rc2/bazel-1.0.0rc2-linux-x86_64...
INFO: Writing tracer profile to '/home/jenkins/.cache/bazel/_bazel_jenkins/4e8644684552b40c50dc624b79e09982/command.profile.gz'
INFO: Invocation ID: 3e92d1fa-39a5-4266-a806-642073c70b89
INFO: Analyzed target //java/com/google/gerrit/server:server (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
Target //java/com/google/gerrit/server:server up-to-date:
  bazel-bin/java/com/google/gerrit/server/libserver.jar
INFO: Elapsed time: 0.625s, Critical Path: 0.21s
INFO: 0 processes.
INFO: Build completed successfully, 1 total action
```

And now yet additional Bazel version was downloaded by Bazelisk and is located here:

```
/home/davido/.cache/bazelisk/bin/bazelbuild/bazel-1.0.0rc2-linux-x86_64
```

PR #93, opened 2019-10-11
Allow using local Bazel binaries.

philwo:
Bazel versions (e.g. in USE_BAZEL_VERSION or .bazelversion) can now
refer to absolute paths on the filesystem. As an added convenience, a
tilde prefix is expanded to the user's home directory.

Example:

```sh
$ USE_BAZEL_VERSION="~/bin/bazel-1.0" bazelisk
```

This would tell Bazelisk to not download any binaries and instead
just directly use the Bazel binary in `$HOME/bin/bazel-1.0`.

### DIFF
--- a/bazelisk.go
+++ b/bazelisk.go
@@ -35,6 +35,7 @@ import (
 	"time"
 
 	version "github.com/hashicorp/go-version"
+	homedir "github.com/mitchellh/go-homedir"
 )
 
 const (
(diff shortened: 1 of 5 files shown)


bazelisk.go
```diff
@@ -686,30 +687,44 @@ func main() {
...
+		bazelDirectory := filepath.Join(bazeliskHome, "bin", bazelFork)
+		err = os.MkdirAll(bazelDirectory, 0755)
+		if err != nil {
+			log.Fatalf("could not create directory %s: %v", bazelDirectory, err)
+		}
+
+		bazelPath, err = downloadBazel(bazelFork, resolvedBazelVersion, isCommit, bazelDirectory)
+		if err != nil {
+			log.Fatalf("could not download Bazel: %v", err)
+		}
```
philwo: (It's hard to see in the diff, but this is just the old code indented one more block.)

davido:
Would this help to solve #89?

philwo:
> Would this help to solve #89?

Not yet - we don't have version detection for Bazel from local binaries, so Bazelisk does not know which Bazel version that binary actually is that you point it to. This in turn means that it can't automatically use a binary that's present on the system if it happens to be the same version that it would download.

Issue #6, opened 2018-10-29
Interaction with pyenv is breaking


natpen:
I'm not sure if this will be controversial as an issue, but I wanted to share a challenge I've faced trying to get Bazelisk working for my codebase. It has a lot to do with the lack of interpreter hermeticity in the Python rules, which is a separate problem, so this "issue" may also resolve itself in the future.

So, what happens is: if you manage your Python by way of [pyenv](https://github.com/pyenv/pyenv), the hashbang in this script (`#!/usr/bin/env python3`) will resolve to the pyenv shim for python 3. This will prepend things to PATH, and spawned subprocess will inherit the mutated PATH as _its_ environment. Further usage of python after bazelisk hands off control, then, end up getting forced onto python 3.

You could absolutely say this not Bazelisk's problem, but for what's it's worth, I worked around this by refactoring the script to be py2-compatible, and using `#!/usr/bin/python` as the hashbang - this should be portable across OSX and Linux, for the time being, and I'm hopeful the Bazel Python ecosystem will continue to improve so this is not something we have to worry about outside of the python rules themselves.

philwo:
> You could absolutely say this not Bazelisk's problem, but for what's it's worth, I worked around this by refactoring the script to be py2-compatible, and using #!/usr/bin/python as the hashbang

This sounds like a good idea. I'd be happy to merge a PR that makes Bazelisk compatible with Python 2 and 3 at the same time. :)

philwo:
Bazelisk (Python version) is now compatible with Python 2 and 3. I'm adding a test that ensures that this continues to be the case.

For the hashbang line, I'm not sure what to do. It seems like there's just no good way to ensure that it works on all systems: https://blogs.gnome.org/mcatanzaro/2018/02/16/on-python-shebangs/

I'd say - feel free to replace it with whatever fits your use case best.

I'll leave it like this for now, also because Bazel's way of finding Python installations on the system [has changed a lot in the last release](https://github.com/bazelbuild/bazel/issues/7899) and the Go version of Bazelisk is gaining more traction and features. Hope that's OK with you!

PR #98, opened 2019-10-30
Fix path comparison between tools/bazel and bazelisk source

clintharrison:
The `os.path.join()` here was resulting in `~/foo/./tools/bazel` being compared to `~/foo/tools/bazel`, which should actually match, but didn't.

### DIFF
--- a/bazelisk.py
+++ b/bazelisk.py
@@ -298,8 +298,13 @@ def delegate_tools_bazel(bazel_path):
     if root:
         wrapper = os.path.join(root, TOOLS_BAZEL_PATH)
         if os.path.exists(wrapper) and os.access(wrapper, os.X_OK):
-            if wrapper != os.path.abspath(__file__):
-                return wrapper
+            try:
+                if not os.path.samefile(wrapper, __file__):
+                    return wrapper
+            except AttributeError:
+                # Python 2 on Windows does not support os.path.samefile
+                if os.path.abspath(wrapper) != os.path.abspath(__file__):
+                    return wrapper
     return None
 
 


philwo:
@clintharrison Alternatively, what about using `os.path.samefile` here?

> Return True if both pathname arguments refer to the same file or directory. This is determined by the device number and i-node number and raises an exception if an os.stat() call on either pathname fails.

clintharrison:
> @clintharrison Alternatively, what about using `os.path.samefile` here?
> 
> > Return True if both pathname arguments refer to the same file or directory. This is determined by the device number and i-node number and raises an exception if an os.stat() call on either pathname fails.

Cool! I like that more -- I didn't realize py3 supported that on Windows now. It's not Windows-compatible in py2, but it looks like we don't intend on having py2 compatibility anymore?

philwo:
Mhm, I would like to preserve Python 2.7 compatibility until the end of this year. What about this:

```python
try:
  if not os.path.samefile(wrapper, __file__):
    return wrapper
except AttributeError:
  # Python 2.x on Windows does not support os.path.samefile yet.
  if os.path.abspath(wrapper) != os.path.abspath(__file__):
    return wrapper
```

We can then remove the try/catch when we finally drop Python 2.x support together with everyone else in the world beginning of 2020 😊

clintharrison:
Ok, that's fair! That looks like a good alternative to me.

(Though I do wonder if anyone runs it with py2 in practice? It has a py3 shebang.)

philwo:
> Ok, that's fair! That looks like a good alternative to me.
> 
> (Though I do wonder if anyone runs it with py2 in practice? It has a py3 shebang.)

You have a point there :D I just checked the issue history and found that I actually promised users in https://github.com/bazelbuild/bazelisk/issues/6#issuecomment-503432216 that it's compatible with Python 2 and 3. I don't remember why I didn't change the shebang to just /usr/bin/python at the time. 🤔

However, luckily this will all be much simpler in 2 months, when we can move to Python 3 only.

PR #107, opened 2020-01-03
add support to providing custom url for downloading the binaries

ali5h:
this is usefull to have for local networks. As long as the provided
url follows github release page format. For example, if you use
an artifactory repository and point it to github.

How it works is one can set `BAZELISK_BASE_URL` as long as that url follows the same pattern as github release page.

---
This change is [Reviewable](https://reviewable.io/reviews/bazelbuild/bazelisk/107)

### DIFF
--- a/bazelisk.go
+++ b/bazelisk.go
@@ -43,7 +43,6 @@ const (
 	skipWrapperEnv = "BAZELISK_SKIP_WRAPPER"
 	wrapperPath    = "./tools/bazel"
 	bazelUpstream  = "bazelbuild"
-	githubUpstream = "https://github.com/%s/bazel/releases/download"
 )
 
 var (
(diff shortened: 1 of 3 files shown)


ali5h:
@philwo would you take a look

bazelisk.go
```diff
@@ -388,6 +389,11 @@ func determineURL(fork string, version string, isCommit bool, filename string) s
 		return fmt.Sprintf("https://storage.googleapis.com/bazel-builds/artifacts/%s/%s/bazel", platforms[runtime.GOOS], version)
```
philwo: I think when the user specifies a local mirror, we should also redirect these nightly builds to that server or alternatively just not support them and error out.

bazelisk.go
```diff
@@ -396,11 +402,16 @@ func determineURL(fork string, version string, isCommit bool, filename string) s
 		kind = "rc" + versionComponents[1]
 	}
 
+	if baseURL != githubUpstream {
```
philwo: It seems like we could simplify this:
- Remove the global constant `githubUpstream`.
- Remove the above check (L392 - 396).
- Move `baseURL := os.Getenv("BAZELISK_BASE_URL")` to above this line.
- Change this check to the above: `if len(baseURL) == 0`.
- Revert below to the old code.

WDYT?

bazelisk.go
ali5h: this mean that i need to move `os.Getenv("BAZELISK_BASE_URL")` to above this `isCommit` check

bazelisk.go
ali5h: i did it a bit different to support commit based downloads too, plz check it out

philwo:
This is a really nice and clear solution! Thanks a lot :)

PR #109, opened 2020-01-13
Split main() to get_bazel_path() and execute_bazel()

sbalabanov:
Some tooling may want to call Bazel more than once per invocation still using Bazelisk. They can integrate with Bazelisk by calling main() but that would mean a potential performance hit for Bazelisk to check Bazel version and cached binary multiple times.
With this change, Python tooling can import `get_bazel_path()` and `execute_bazel()` from bazelisk.py and call them appropriately.

### DIFF
--- a/bazelisk.py
+++ b/bazelisk.py
@@ -325,10 +325,7 @@ def execute_bazel(bazel_path, argv):
             pass
 
 
-def main(argv=None):
-    if argv is None:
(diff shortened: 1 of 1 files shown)


sbalabanov:
@googlebot I signed it!

sbalabanov:
@philwo Done, thanks!

PR #114, opened 2020-01-31
Add an npm package output

alexeagle:
/cc @achew22

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -1,5 +1,6 @@
 load("@io_bazel_rules_go//go:def.bzl", "go_binary", "go_library")
 load("@bazel_gazelle//:def.bzl", "gazelle")
+load("@build_bazel_rules_nodejs//:index.bzl", "pkg_npm")
 
 # gazelle:prefix github.com/bazelbuild/bazelisk
 gazelle(name = "gazelle")
@@ -53,4 +54,49 @@ go_binary(
     name = "bazelisk",
     embed = [":go_default_library"],
     visibility = ["//visibility:public"],
+)
+
+
+go_binary(
+    name = "bazelisk-darwin",
+    out = "bazelisk-darwin_amd64",
+    embed = [":go_default_library"],
+    goarch = "amd64",
+    goos = "darwin",
+    pure = "on",
+    visibility = ["//visibility:public"],
+)
+
+go_binary(
+    name = "bazelisk-linux",
+    out = "bazelisk-linux_amd64",
+    embed = [":go_default_library"],
+    goarch = "amd64",
+    goos = "linux",
+    pure = "on",
+    visibility = ["//visibility:public"],
+)
+
+go_binary(
+    name = "bazelisk-windows",
+    out = "bazelisk-windows_amd64.exe",
+    embed = [":go_default_library"],
+    goarch = "amd64",
+    goos = "windows",
+    pure = "on",
+    visibility = ["//visibility:public"],
+)
+
+pkg_npm(
+    name = "npm_package",
+    srcs = [
+        "LICENSE",
+        "package.json",
+        "bazelisk.js",
+    ],
+    deps = [
+        ":bazelisk-darwin",
+        ":bazelisk-linux",
+        ":bazelisk-windows",
+    ],
 )
--- a/bazelisk.js
+++ b/bazelisk.js
@@ -1,0 +1,73 @@
+#!/usr/bin/env node
+// Copyright 2020 The Bazel Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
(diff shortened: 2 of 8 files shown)


BUILD
```diff
@@ -54,3 +55,48 @@ go_binary(
     embed = [":go_default_library"],
     visibility = ["//visibility:public"],
 )
+
+
+go_binary(
+    name = "bazelisk-darwin",
+    out = "bazelisk-darwin_amd64",
+    embed = [":go_default_library"],
+    goarch = "amd64",
```
philwo: I didn't know that one could create platform-specific go_binary targets - that's much better than my existing approach in `build.sh`! :) Thanks!

bazelisk.js
```diff
@@ -0,0 +1,73 @@
...
+  const extension = {
+    'darwin' : '',
+    'linux' : '',
+    'win32' : '.exe',
+  }[os.platform()];
+
+  if (arch == undefined || platform == undefined) {
+    console.error(`FATAL: Your platform/architecture combination ${
+        os.platform()} - ${os.arch()} is not yet supported.
+    Follow install instructions at https://github.com/bazelbuild/bazel-watcher/blob/master/README.md to compile for your system.`);
```
philwo: How are the bazel-watcher install instructions related to building Bazel on an unsupported architecture? 🤔

bazelisk.js
```diff
@@ -0,0 +1,73 @@
...
+  ps.on('close', e => process.exitCode = e);
+}
+
+if (require.main === module) {
+  main(process.argv.slice(2));
+}
+
+module.exports = {
+  getNativeBinary,
+};
```
philwo: nit: add trailing newline?

build.sh
```diff
@@ -44,3 +41,10 @@ rm -f bazelisk
 ### Print some information about the generated binaries.
 ls -lh bin/*
 file bin/*
+
+# Non-googlers: you should run this script with NPM_REGISTRY=https://registry.npmjs.org
+readonly REGISTRY=${NPM_REGISTRY:-https://wombat-dressing-room.appspot.com}
+# Googlers: you should npm login using the go/npm-publish service:
+# $ npm login --registry https://wombat-dressing-room.appspot.com
+# TODO: also publish the release to npm.
+# ./bazelisk run --config=release //:npm_package.publish -- --access=public --tag latest --registry $REGISTRY
```
philwo: Does the "TODO" mean that this is a manual step, or is there still something missing before we can publish to npm?

package.json
```diff
@@ -0,0 +1,8 @@
+{
+    "name": "@bazel/bazelisk",
+    "version": "0.0.0-PLACEHOLDER",
```
philwo: How is the "version" here replaced with the correct version?

rules_nodejs.patch
```diff
@@ -0,0 +1,13 @@
+diff internal/pkg_npm/packager.js internal/pkg_npm/packager.js
```
philwo: What does the patch do?

BUILD
alexeagle: Yeah I figured you might want to refactor that script too. If nothing else, this allows you to build all the architectures in parallel, even if you still loop through them to place them in a `bin` directory

bazelisk.js
alexeagle: Oops this is some leftover copy-pasta. Fixed

build.sh
alexeagle: It just needs some manual testing. I could do that myself, but maybe it's better for you or Florian to do the first release to make sure permissions are right.

package.json
alexeagle: It's a feature of the pkg_npm rule: https://github.com/bazelbuild/rules_nodejs/blob/master/docs/Built-ins.md#replace_with_version

rules_nodejs.patch
alexeagle: I've never used pkg_npm in a repo that uses `v` prefix on the tags. NPM package versioning has to be purely numeric so it needs to be stripped. I'll make the same change upstream.

build.sh
alexeagle: I removed the comment, but it means after merging this, you'll trip over any problems with npm publishing the next time you run it.

(Also I notice the `build.sh` script doesn't actually do a publish step yet, so maybe this isn't the right place for this command to be added? I didn't see any other release instructions where it belongs instead)

stamp.sh
```diff
@@ -7,3 +7,5 @@ CURRENT_TAG=$(git tag -l --points-at HEAD | head -n1)
 CURRENT_COMMIT=$(git rev-parse HEAD)
 
 echo "STABLE_VERSION ${CURRENT_TAG:-$CURRENT_COMMIT}"
+# rules_nodejs expects to read from volatile-status.txt
+echo "BUILD_SCM_VERSION ${CURRENT_TAG:-$CURRENT_COMMIT}"
```
achew22: I've found it valuable to add a `-dirty` suffix when user changes happen. This saved me trying to debug someone's issue which would have been impossible without it

alexeagle: Agreed, for Angular we use a `.with-local-changes` suffix in a non-pristine clone
https://github.com/angular/angular/blob/master/tools/bazel_stamp_vars.js#L52-L53

But of course that change should be made for all the stamping here, not just npm, so I don't think it should go in this PR. @philwo do you want a separate PR for that?

alexeagle:
Ping! I'd like to build on top of this. If you're okay with it, I could publish the first version of the package from the last release tag.

philwo:
Sorry for the delay, @alexeagle! Merging this now and will try to push a first release.

PR #125, opened 2020-03-10
Add --//some/path:incompatible_some_flag  as a valid incompatible flag format

### LINKED ISSUE #118: Allow Starlark defined incompatible flags (e.g. //some/path:incompatible_some_flag
This feature would unblock https://github.com/bazelbuild/bazel/issues/8857

It looks like this might be as simple as changing the regex here: https://github.com/bazelbuild/bazelisk/blob/4b1375bdada94be345679335a4c32cf179f07711/bazelisk.go#L568

aiuto:
fixes #118


philwo:
Does that mean we have actual flags that start with `--` and then go on with a label syntax in Bazel?

Like this?

```
bazel build --//some/path:incompatible_user_defined_flag --spawn_strategy=foobar //my:app
```

aiuto:
On Tue, Mar 10, 2020 at 8:56 AM Philipp Wollermann <notifications@github.com>
wrote:

> Does that mean we have actual flags that start with -- and then go on
> with a label syntax in Bazel?
>
> Like this?
>
> bazel build --//some/path:incompatible_user_defined_flag --spawn_strategy=foobar //my:app
>
>
Yes.  https://github.com/bazelbuild/bazel/issues/8857

It is removal of starlark build_defs. As we get incompatible changes in
rules we are going to see more of this.

aiuto:
friendly ping

PR #126, opened 2020-03-23
Make bazelisk available as "bazel" when installing via NPM

lewish:
Right now having to install bazelisk and then setup a symlink to it under e.g. `/usr/bin/bazel` is a bit annoying.

By adding this, a user can immediately start running `bazel` commands via bazelisk (as well as `bazelisk` commands) after doing:

```
npm i -g @bazel/bazelisk
```

This is the suggested way to use bazelisk in the readme, and for the nodejs users I struggle to see many scenarios where you wouldn't want this to happen.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,8 @@
     "version": "0.0.0-PLACEHOLDER",
     "license": "Apache-2.0",
     "bin": {
-        "bazelisk": "bazelisk.js"
+        "bazelisk": "bazelisk.js",
+        "bazel": "bazelisk.js"
     },
     "keywords": [
         "bazel"


lewish:
@alexeagle

Toxicable:
I like this change, we've avoided using `bazelisk` since our current advice to devs is to use `npx bazel` \ `npx ibazel` and I don't want to tell them to use `npx bazelisk` \ `npx ibazel`

However, we should sync this change with deprecating the @bazel/bazel package

alexeagle:
I agree this is a better user experience. I avoided including `bazel` in the bin for bazelisk in the initial PR mostly as @Toxicable says, that there's already the `@bazel/bazel` package, which is at version 2.1.0 and will probably never be published again. @josephperrott might remember more of that conversation. (Also Joey, I think I'll need you to help deprecate the `@bazel/bazel` package since I don't have angular npm credentials anymore)

A few other thoughts about it.
- https://www.npmjs.com/package/bazel is 404. If you really want `npx bazel` to be reliable, it needs to have that package name. Otherwise the behavior depends on whether you've already global-installed a package with that bin. If not, then `npx @bazel/bazel` is the reliable instruction to give anyone since it doesn't rely on the state of their machine
- This makes npm users different from other language ecosystems. Here we'll say that "install bazel" == bazelisk features, where other users have to take a separate install step. We ought to try to head in the same direction at least. @philwo do you imagine Bazel team will get to a point where following the standard bazel install instructions actually gives you bazelisk as the binary?
- Bazel and bazelisk are still different tools with slightly different semantics (eg. the latter has a `--strict` flag that doesn't exist on the former). If users actually care to have the distinction between the two of them, it will be harder to understand which one you're running. This might have implications for helping resolve ppl's issues where they think they are running bazel but it's actually bazelisk. That said, I think anyone following the symlink/alias install instructions will forget they did that, and similarly be confused which is which.

josephperrott:
I believe one of the other driving factors in our conversation is that npm doesn't inform you if the same binary is defined in multiple packages.

If you install both `@bazel/bazel` and `@bazel/bazelisk` and they both define a binary, `bazel`, it's not apparent which one will be installed and run locally.  This is even more difficult to determine when they act, for many usages, as the exact same thing.

josephperrott:
Summary of my conversation with @alexeagle offline:

We think that we should:
- deprecate @bazel/bazel on npm with a message instructing migration to @bazel/bazelisk
- update the @bazel/bazel binary script that wraps bazel's actual binary to instruct migration to @bazel/bazelisk

Both of these aren't really a great idea until @bazel/bazelisk has a `bazel` binary, so that its a drop in replacement.

Ideally we can do this in a way that allows us to never have to publish a `3.0` of @bazel/bazel so that its a clean break for users.

philwo:
@alexeagle Do I understand correctly that I can just cut a new Bazelisk release and push that to NPM as a first step?

PR #142, opened 2020-07-04
Prepend directory containing relevant `bazel` binary to `PATH`

### LINKED ISSUE #111: Bazelisk subprocesses should prepend the correct bazel to the $PATH
Many bazel rules end up running shell scripts that invoke `bazel`, which resolves to the first bazel binary in the path.

For example:

https://github.com/bazelbuild/rules_jvm_external/blob/0a7cc6a0b6764232a0ddd31ad87b489e1d47b166/private/pin.sh#L5

This should resolve to the `bazel` that `bazelisk` downloaded, otherwise you can end up using a different `bazel` binary in these subcommands. This causes confusing errors.

I believe the most natural fix would be for `bazelisk` to prepend a directory containing the relevant `bazel` binary to the `$PATH`.

coeuvre:
Fixes #111

### DIFF
--- a/bazelisk.go
+++ b/bazelisk.go
@@ -579,7 +579,7 @@ func downloadBazel(fork string, version string, isCommit bool, directory string)
 	return destinationPath, nil
 }
 
-func copyFile(src, dst string, perm os.FileMode) (err error) {
+func copyFile(src, dst string, perm os.FileMode) error {
 	srcFile, err := os.Open(src)
 	if err != nil {
 		return err
--- a/bazelisk.go
+++ b/bazelisk.go
@@ -580,16 +580,26 @@ func downloadBazel(fork string, version string, isCommit bool, directory string)
 }
 
 func copyFile(src, dst string, perm os.FileMode) (err error) {
-	data, err := ioutil.ReadFile(src)
+	srcFile, err := os.Open(src)
 	if err != nil {
 		return err
 	}
-	return ioutil.WriteFile(dst, data, perm)
+	defer srcFile.Close()
+
+	dstFile, err := os.OpenFile(dst, os.O_WRONLY|os.O_CREATE, perm)
+	if err != nil {
+		return err
+	}
+	defer dstFile.Close()
+
+	_, err = io.Copy(dstFile, srcFile)
+
+	return err
 }
 
 func linkLocalBazel(directory string, bazelPath string) (string, error) {
-	directoryName := dirForURL(bazelPath)
-	destinationDir := filepath.Join(directory, directoryName, "bin")
+	normalizedBazelPath := dirForURL(bazelPath)
+	destinationDir := filepath.Join(directory, normalizedBazelPath, "bin")
 	err := os.MkdirAll(destinationDir, 0755)
 	if err != nil {
 		return "", fmt.Errorf("could not create directory %s: %v", destinationDir, err)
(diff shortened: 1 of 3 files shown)


philwo:
Thanks Chi! Assigning to @fweikert for review.

bazelisk.go
```diff
@@ -889,15 +956,16 @@ func main() {
...
+			log.Fatalf("could not download Bazel: %v", err)
 		}
-
-		bazelPath, err = downloadBazel(bazelFork, resolvedBazelVersion, isCommit, bazelDirectory)
+	} else {
+		baseDirectory := filepath.Join(bazeliskHome, "local")
+		bazelPath, err = linkLocalBazel(baseDirectory, bazelPath)
 		if err != nil {
-			log.Fatalf("could not download Bazel: %v", err)
+			log.Fatalf("cound not link loal Bazel: %v", err)
```
fweikert, line 968: local

bazelisk.go
```diff
@@ -566,6 +579,41 @@ func downloadBazel(fork string, version string, isCommit bool, directory string)
...
+	data, err := ioutil.ReadFile(src)
+	if err != nil {
+		return err
+	}
+	return ioutil.WriteFile(dst, data, perm)
+}
+
+func linkLocalBazel(directory string, bazelPath string) (string, error) {
+	directoryName := dirForURL(bazelPath)
+	destinationDir := filepath.Join(directory, directoryName, "bin")
```
fweikert, line 592: please choose more a more descriptive name for at least one of directory / directoryName

bazelisk.go
```diff
@@ -566,6 +579,41 @@ func downloadBazel(fork string, version string, isCommit bool, directory string)
...
+		if runtime.GOOS == "windows" {
+			// If can't create Symlink on Windows, fallback to copy
+			if err != nil {
+				err = copyFile(bazelPath, destinationPath, 0755)
+				if err != nil {
+					return "", fmt.Errorf("cound not copy file from %s to %s: %v", bazelPath, destinationPath, err)
+				}
+			}
+		} else {
+			if err != nil {
```
fweikert, line 609: We could also use the fallback on other platforms.

bazelisk.go
```diff
@@ -585,6 +633,24 @@ func maybeDelegateToWrapper(bazel string) string {
...
+func prependDirToPathList(cmd *exec.Cmd, dir string) {
+	found := false
+	for idx, val := range cmd.Env {
+		splits := strings.Split(val, "=")
+		if len(splits) != 2 {
+			continue
+		}
+		if splits[0] == "PATH" {
+			found = true
+			cmd.Env[idx] = fmt.Sprintf("PATH=%s%s%s", dir, string(os.PathListSeparator), splits[1])
```
fweikert, line 645: You can "break" in this case

bazelisk.go
```diff
@@ -566,6 +579,41 @@ func downloadBazel(fork string, version string, isCommit bool, directory string)
 	return destinationPath, nil
 }
 
+func copyFile(src, dst string, perm os.FileMode) (err error) {
+	data, err := ioutil.ReadFile(src)
+	if err != nil {
+		return err
+	}
+	return ioutil.WriteFile(dst, data, perm)
```
philwo, line 587: I would prefer to use `io.Copy` here instead of `ioutil.ReadFile` and `ioutil.WriteFile`, so that we don't read the entire file into RAM first, as the Bazel binary can be quite large.

bazelisk.go
philwo: +1. First I wanted to say that symlinks always work on non-Windows platforms, but then I remembered the time where someone tried to run Bazel on Linux with $HOME being on a `vfat` volume...

I think it's fine to just always fallback to copy.

philwo:
Could you add a test for this to `bazelisk_test.sh`?

coeuvre:
Updated and added test cases to `bazelisk_test.sh`.

bazelisk.go
```diff
@@ -566,6 +579,45 @@ func downloadBazel(fork string, version string, isCommit bool, directory string)
 	return destinationPath, nil
 }
 
+func copyFile(src, dst string, perm os.FileMode) (err error) {
```
fweikert, line 582: Why is this a named return parameter? This feature is not being used in the function body (i.e. there aren't any bare return statements).

coeuvre: You are right. The named return parameter is unnecessary.

bazelisk.go
fweikert: This is the only remaining comment from my side. Other than that: LGTM, great work!

PR #148, opened 2020-08-11
Also check if `tools/bazel` is a directory

Kiougar:
If `tools/bazel` is a directory, the `maybeDelegateToWrapper` function tries to execute it, which results in a permission denied error.

This commit prevents that by checking if `tools/bazel` is a directory and ignores it.


philwo:
Thanks!

PR #152, opened 2020-08-18
Fix a bug that PATH on Windows will sometimes get overwritten

### LINKED ISSUE #149: Bazelisk 1.6.0 overwrites PATH
I was writing some repository rules and I just couldn't get why the ```PATH``` variable stored only the path to the Bazel binary, no matter what combination of ```--action_env```s or ```--repo_env```s I used (and now I know they don't even matter in the case of repository rules and ```PATH```). 

After some time of feeling hopeless, my mind was struck by the idea of executing Bazel directly, not by Bazelisk. And... this time```PATH``` was the same as the system one. That gave me just a little bit of hope that I'm not actually going insane.

So after that, I decided to try the previous version of Bazelisk, ```1.5.0```, and ```PATH``` was also the same as the system one.

Patch notes for ```1.6.0``` say: 

> Bazelisk now runs Bazel with a $PATH set so that running the bazel command resolves to the correct Bazel version. For example, this is useful if you have shell scripts or tests that invoke bazel - these will now pick the right version of Bazel. Thanks for the contribution @coeuvre! (#111)

and #111 says:

> Bazelisk subprocesses should prepend the correct bazel to the $PATH

So it seems that instead of prepending, Bazelisk is overwriting ```PATH```. Or I'm misunderstanding something.

I'm on Windows btw.

coeuvre:
Fixes #149

### DIFF
--- a/bazelisk.go
+++ b/bazelisk.go
@@ -646,7 +646,7 @@ func prependDirToPathList(cmd *exec.Cmd, dir string) {
 		if len(splits) != 2 {
 			continue
 		}
-		if splits[0] == "PATH" {
+		if strings.ToUpper(splits[0]) == "PATH" {
 			found = true
 			cmd.Env[idx] = fmt.Sprintf("PATH=%s%s%s", dir, string(os.PathListSeparator), splits[1])
 			break
--- a/bazelisk.go
+++ b/bazelisk.go
@@ -646,7 +646,7 @@ func prependDirToPathList(cmd *exec.Cmd, dir string) {
 		if len(splits) != 2 {
 			continue
 		}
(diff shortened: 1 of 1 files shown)


coeuvre:
@philwo

philwo:
As a related discussion recently came up on HackerNews (https://news.ycombinator.com/item?id=24177472), I would suggest to use a case-insensitive string comparison method rather than `ToUpper` or `ToLower`.

This one seems good (and is probably even faster): https://golang.org/pkg/strings/#EqualFold

Not that it matters for this specific comparison, but we can set a good example here. ;)

coeuvre:
Updated. Thanks for suggestion!

PR #157, opened 2020-09-01
Fix dispatch with no arguments in bazelisk.py

### LINKED ISSUE #156: `bazelisk.py` fails with exception when called without arguments.
`bazelisk.py` failed with an unhandled exception when called without arguments. The expected behavior would be to mirror the go binary behavior and invoke `bazel` without arguments.

```
$ ./bazelisk.py
Traceback (most recent call last):
  File "./bazelisk.py", line 406, in <module>
    sys.exit(main())
  File "./bazelisk.py", line 395, in main
    if argv[0] == "--print_env":
IndexError: list index out of range
```

asford:
fixes #156

### DIFF
--- a/bazelisk.py
+++ b/bazelisk.py
@@ -392,7 +392,7 @@ def main(argv=None):
 
     argv = argv[1:]
 
-    if argv[0] == "--print_env":
+    if argv and argv[0] == "--print_env":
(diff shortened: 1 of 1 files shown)


philwo (approved):
Thank you!

PR #159, opened 2020-09-03
refresh rules deps and bump go to v1.15

chenrui333:


meteorcloudy:
This PR broke Bazelisk test on Bazel CI:
https://buildkite.com/bazel/bazelisk/builds/741

chenrui333:
might be permission issue 
```

[Errno 30] Read-only file system: '/go_bazelisk_test'
--
  | buildkite-agent artifact upload
  | 2020-09-03 08:07:42 FATAL  Missing upload paths. See: `buildkite-agent artifact upload --help`
```

meteorcloudy:
That one is also weird. But I think the failure of `//:go_bazelisk_test` is also valid. Can you reproduce the failure locally?

fweikert:
Post-submit is still red :(

philwo:
```
Could not find the bazelisk executable, listing files:
--
  | .
  | ./WORKSPACE
  | ./BUILD
  | ./log
```

The issue is that the test framework does not copy the Bazelisk binary into the working directory, because the glob doesn’t match it. I thought I had fixed this already? 🧐

fweikert:
For some reason the Go rules changed the output layout. The test passes when I change the binary path to

`$(rlocation __main__/bazelisk_/bazelisk)`

which points at /private/var/tmp/_bazel_fwe/ba2df144dcb48aef9b03bc0348b3fd38/sandbox/darwin-sandbox/28/execroot/__main__/bazel-out/darwin-fastbuild/bin/go_bazelisk_test.runfiles/__main__/bazelisk_/bazelisk in my case (MacBook).

Issue #171, opened 2020-09-27
Bazelisk-Go: Version history from GCS is incorrect wrt release candidates


fweikert:
Problem:

When setting the Bazel version to `latest`, `latest-1` etc. or no value at all, Bazelisk-go cannot handle Bazel versions that have a release candidate, but no release yet. 

Explanation:

This is a bug in the code that retrieves available Bazel versions from GCS: The code currently fetches the names of all available directories in the bucket (e.g. '3.4.0', '3.4.1', '3.5.0', ...), then sorts them and checks whether the last element has a release/ subdirectory. If not, the last element is removed from the version history before returning the shorter list.

There are two problems with this approach:
1. `latest` does not work if both the last version and the second-to-last one don't have a full release yet, since the second-to-last version is never examined in detail (example breakage: #170).
2. `latest-n` returns an incorrect result if any of the latest N versions (except the very last one) does not have a release yet.

Problem 1 can be fixed easily by looping until the last version has a release. Problem 2 is harder to fix since we would have to check *every* existing version for whether a corresponding release exists (i.e. starting with 0.1.0). This change would have severe performance implications, which is especially bad since this code path is executed very frequently. One idea would be to change the `GetReleaseVersions()` function of the `ReleaseRepo` interface by adding an additional "lastN" parameter, which would reduce the number of GCS queries.

PR #172, opened 2020-09-27
GO: Return latest version from GCS that is a full release

### LINKED ISSUE #170: Bazelisk is pulling a non existing release ( 3.5.1 ) when invoked w/o .bazelversion
Invoking `bazelisk build` is attempting to pull a non-existent release.

This afternoon, a handful of xchammer github jobs started failing that used bazelisk. For example:
https://github.com/pinterest/xchammer/pull/259/checks?check_run_id=1168118720

The xchammer CI runs `bazelisk` and we don't have a version specified in the repository. Without the version, it attempts to pull a non-existing URL:
```
2020/09/25 21:40:42 Downloading https://releases.bazel.build/3.5.1/release/bazel-3.5.1-darwin-x86_64...
2020/09/25 21:40:42 could not download Bazel: HTTP GET https://releases.bazel.build/3.5.1/release/bazel-3.5.1-darwin-x86_64 failed with error 404
make: *** [build-impl] Error 1
Error: Process completed with exit code 2.```

I checked on Bazel releases this afternoon, and it looks like there is no `3.5.1` bazel release: https://github.com/bazelbuild/bazel/releases

### PR DESCRIPTION
**Pull request by fweikert**
If `latest-n` was requested, Bazelisk previously only examined the most recent version to see whether it had a release (and not just candidates).
If there was no release, Bazelisk simply returned the second-to-last one without checking whether that version had a release, too.

Now Bazelisk checks the latest versions until it finds one with a release.

Fixes #170
Progress towards #171

### DIFF
--- a/repositories/gcs.go
+++ b/repositories/gcs.go
@@ -49,14 +49,22 @@ func getVersionHistoryFromGCS(onlyFullReleases bool) ([]string, error) {
 	available := getVersionsFromGCSPrefixes(prefixes)
 	sorted := versions.GetInAscendingOrder(available)
 
-	if onlyFullReleases && len(sorted) > 0 {
-		latestVersion := sorted[len(sorted)-1]
-		_, isRelease, err := listDirectoriesInReleaseBucket(latestVersion + "/release/")
-		if err != nil {
-			return []string{}, fmt.Errorf("could not list release candidates for latest release: %v", err)
+	// TODO(#171): This algorithm is incorrect if version == 'latest-n' and any of the last n versions (except the last one) does not have a release yet.
+	if onlyFullReleases {
+		for len(sorted) > 0 {
+			lastIndex := len(sorted)-1
+			latestVersion := sorted[lastIndex]
+			_, isRelease, err := listDirectoriesInReleaseBucket(latestVersion + "/release/")
+			if err != nil {
+				return []string{}, fmt.Errorf("could not list release candidates for latest release: %v", err)
+			}
+			if isRelease {
+				break
+			}
+			sorted = sorted[:lastIndex]
 		}
-		if !isRelease {
-			sorted = sorted[:len(sorted)-1]
+		if len(sorted) == 0 {
+			return []string{}, errors.New("there are no releases available")
 		}
 	}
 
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: repositories/gcs.go [hunk 1]
```diff
@@ -49,14 +49,22 @@ func getVersionHistoryFromGCS(onlyFullReleases bool) ([]string, error) {
...
-		if err != nil {
-			return []string{}, fmt.Errorf("could not list release candidates for latest release: %v", err)
+	// TODO(#171): This algorithm is incorrect if version == 'latest-n' and any of the last n versions (except the last one) does not have a release yet.
+	if onlyFullReleases {
+		for len(sorted) > 0 {
+			lastIndex := len(sorted)-1
+			latestVersion := sorted[lastIndex]
+			_, isRelease, err := listDirectoriesInReleaseBucket(latestVersion + "/release/")
+			if err != nil {
+				return []string{}, fmt.Errorf("could not list release candidates for latest release: %v", err)
```
philwo, line 59: It's not the "latest" release, as we iterate over the array, right? What about this?

```go
return []string{}, fmt.Errorf("could not list release candidates for %v: %v", latestVersion, err)
```

repositories/gcs.go
```diff
@@ -49,14 +49,22 @@ func getVersionHistoryFromGCS(onlyFullReleases bool) ([]string, error) {
 	available := getVersionsFromGCSPrefixes(prefixes)
 	sorted := versions.GetInAscendingOrder(available)
 
-	if onlyFullReleases && len(sorted) > 0 {
-		latestVersion := sorted[len(sorted)-1]
-		_, isRelease, err := listDirectoriesInReleaseBucket(latestVersion + "/release/")
-		if err != nil {
-			return []string{}, fmt.Errorf("could not list release candidates for latest release: %v", err)
+	// TODO(#171): This algorithm is incorrect if version == 'latest-n' and any of the last n versions (except the last one) does not have a release yet.
```
philwo, line 52: Just for my understanding, what we would actually want to do here is to basically filter the `sorted` array and remove all entries for which no release version exists, right? But we don't do that, because you think it's too slow due to the number of HTTP requests?

philwo (approved):

repositories/gcs.go
fweikert: Yes, that's the problem. But we can work around this by specifying how many of the latest releases we need: #171

PR #199, opened 2020-11-18
deps: add dependabot

chenrui333:
add dependabot for keeping the dependencies up to date.


philwo:
Hey, that's cool! Thank you :)

PR #205, opened 2020-12-17
Supporting BAZELISK_BASE_URL in Python version

linzhp:
The environment variable `BAZELISK_BASE_URL` is supported in the Go version. I would like to use that in the Python version too.


philwo (approved):
Thank you! And sorry for the late review.

PR #218, opened 2021-03-04
Fix workspace root detection by ignoring directories

avidal:
On case-insensitive filesystems, bazelisk can fail to find the workspace root if you have a directory named `workspace/`. This patch updates `findWorkspaceRoot` to ensure that a workspace root is only returned if the WORKSPACE / WORKSPACE.bazel file exists and is not a directory, and makes bazelisk consistent with bazel itself (see comments in the patch for a link)

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -200,12 +200,24 @@ func GetEnvOrConfig(name string) string {
 	return fileConfig[name]
 }
 
+// isValidWorkspace returns true iff the supplied path is the workspace root, defined by the presence of
+// a file named WORKSPACE or WORKSPACE.bazel
+// see https://github.com/bazelbuild/bazel/blob/8346ea4cfdd9fbd170d51a528fee26f912dad2d5/src/main/cpp/workspace_layout.cc#L37
+func isWorkspaceRoot(path string) bool {
+	info, err := os.Stat(path)
+	if err != nil {
+		return false
+	}
+	
+	return !info.IsDir()
+}
+
 func findWorkspaceRoot(root string) string {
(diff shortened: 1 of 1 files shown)


core/core.go
```diff
@@ -200,12 +200,24 @@ func GetEnvOrConfig(name string) string {
 	return fileConfig[name]
 }
 
+// isValidWorkspace returns true iff the supplied path is the workspace root, defined by the presence of
+// a file named WORKSPACE or WORKSPACE.bazel
+// see https://github.com/bazelbuild/bazel/blob/8346ea4cfdd9fbd170d51a528fee26f912dad2d5/src/main/cpp/workspace_layout.cc#L37
+func isWorkspaceRoot(path string) bool {
```
philwo: Either this function has to be called `isValidWorkspace` or the function calls below have to be changed to `isWorkspaceRoot`.

philwo (changes requested):

philwo:
Thank you @avidal, this is great!

core/core.go
avidal: Yeah, just fixed it. I edited through github so I didn't see that I forgot to rename the function after changing the call sites.

avidal:
> Thank you @avidal, this is great!

Took a few hours of pulling my hair out wondering why I could force the bazel version via `USE_BAZEL_VERSION` but not using the `.bazelversion` in my workspace root. Eventually I looked at the code and then realized I was on a case-insensitive filesystem :)

philwo (approved):

PR #238, opened 2021-05-05
Properly handle pagination when GCS doesn't give us full results.

meisterT:


meisterT:
cc @fweikert @philwo

fweikert:
Fixes https://github.com/bazelbuild/bazelisk/issues/239

repositories/gcs.go
```diff
@@ -63,27 +63,44 @@ func getVersionHistoryFromGCS() ([]string, error) {
...
-	content, err := httputil.ReadRemoteFile(url, "")
-	if err != nil {
-		return nil, false, fmt.Errorf("could not list GCS objects at %s: %v", url, err)
+		baseURL = fmt.Sprintf("%s&prefix=%s", baseURL, prefix)
 	}
 
-	var response GcsListResponse
-	if err := json.Unmarshal(content, &response); err != nil {
-		return nil, false, fmt.Errorf("could not parse GCS index JSON: %v", err)
+	var result GcsListResponse
```
fweikert, line 71: You could replace this with a []string and an int

fweikert (approved):

repositories/gcs.go
meisterT: Done, also fixed a mistake. PTAL.

fweikert (approved):

fweikert:
A unit test would be nice, too, but for now I would be happy enough if this has been tested manually.

meisterT:
> A unit test would be nice, too, but for now I would be happy enough if this has been tested manually.

Yeah, I tested manually by adding `&maxResults=10` to the URL and observing the results.

PR #244, opened 2021-06-02
Add note about ensuring developers have Bazelisk rather than Bazel

alexeagle:
Also move install instructions to the top, where impatient users will expect those.


philwo (approved):
Thanks!

PR #258, opened 2021-07-21
Use unsigned integer for shift count type

changlan:
https://github.com/bazelbuild/bazelisk/pull/256 introduced an incompatible issue with Golang versions prior to 1.13:

```
github.com/bazelbuild/bazelisk/httputil/httputil.go:117:25: invalid operation: 1 << attempt (shift count type int, must be unsigned integer)
```

This PR fixes the issue by forcing the `attempt` variable as an unsigned integer.


changlan:
gentle ping :)

philwo:
@fweikert WDYT?

fweikert:
Can you please add the casts to the two lines with the shift? Then we can leave the other lines as-is.

changlan:
Done.

fweikert (approved):

fweikert:
Thank you!

PR #262, opened 2021-08-12
suggest 'go install' instead of 'go get'

dmitris:
https://golang.org/doc/go-get-install-deprecation - "Deprecation of 'go get' for installing executables" says that using `go get` for installing binaries is deprecated and will not be supported starting from go1.18. It would be good to get users to get used to the "new way" in advance.


philwo (approved):
Thank you!

PR #268, opened 2021-09-10
Allowing Bazelisk to download Bazel without JDK

### LINKED ISSUE #267: bazel nojdk version
We don't use java at all and would like to use the nojdk version of bazel. (less to download in our ci/cd jobs).

Is it possible to add support for it? Does it actually support it now and I just have not figured it out?

linzhp:
Fixes #267


philwo (approved):
Thank you!

PR #271, opened 2021-09-24
build: add typings/typescript support for @bazel/bazelisk npm package

josephperrott:
Adding a d.ts file expressing the exported `getNativeBinary` function and adding the typings and main attributes to package.json allows for standard typescript importing.  This allows for usage in typescript projects to obtain the path to the bazel native binary to called within the users own script.


philwo:
Sorry for the delay and thanks for the contribution! It seems fine to me.

@alexeagle Any thoughts / worries?

PR #272, opened 2021-10-06
Added support for Apple Silicon

### LINKED ISSUE #270: bazelisk.py does not support Apple arm64
This check here:
https://github.com/bazelbuild/bazelisk/blob/280b07a854a70f2ab0887d644e364c8e8c1c6187/bazelisk.py#L210

prevents it from working on Apple arm64.

The simplest fix is to just delete that check.  If the architecture is not supported, the file will fail to download.

linzhp:
Bazelish should check the Bazel version. If it is greater or equal to than 4.1, add support for arm64 so we can run Bazel on Apple Sillicon. Fixed #270

### DIFF
--- a/bazelisk.py
+++ b/bazelisk.py
@@ -207,6 +207,23 @@ def determine_executable_filename_suffix():
 
 def determine_bazel_filename(version):
     operating_system = get_operating_system()
+    supported_machines = get_supported_machine_archs(version, operating_system)
+    machine = normalized_machine_arch_name()
+    if machine not in supported_machines:
+        raise Exception(
+            'Unsupported machine architecture "{}". Bazel {} only supports {} on {}.'.format(
+                machine, version, ", ".join(supported_machines), operating_system.capitalize()
+            )
+        )
+
+    filename_suffix = determine_executable_filename_suffix()
+    bazel_flavor = "bazel"
+    if os.environ.get("BAZELISK_NOJDK", "0") != "0":
+        bazel_flavor = "bazel_nojdk"
+    return "{}-{}-{}-{}{}".format(bazel_flavor, version, operating_system, machine, filename_suffix)
+
+
+def get_supported_machine_archs(version, operating_system):
     supported_machines = ["x86_64"]
     versions = version.split(".")[:2]
     if len(versions) == 2:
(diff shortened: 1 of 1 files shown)


mindaugasrukas:
LGTM. Works on my M1 machine.

linzhp:
ping @philwo

philwo (approved):
Thank you!

bazelisk.py
```diff
@@ -206,11 +206,22 @@ def determine_executable_filename_suffix():
 
 
 def determine_bazel_filename(version):
+    supported_machines = ["x86_64"]
+    versions = version.split('.')[:2]
+    if len(versions) == 2:
+        # released version
+        major, minor = int(versions[0]), int(versions[1])
+        if major > 4 or major == 4 and minor >= 1:
+            # Apple Sillicon was supported since 4.1:
```
philwo, line 215: nit: s/Sillicon/Silicon/

philwo (dismissed):
Sorry, found two little things to change - could you have a look?

linzhp:
@fweikert I believe the current CI errors are transient issues. It can be fixed by restarting the job, but I don't have the permission to do so.

bazelisk.py
```diff
@@ -206,16 +206,36 @@ def determine_executable_filename_suffix():
...
+            or operating_system == "linux"
+            and (major > 3 or major == 3 and minor >= 4)
+        ):
+            # Linux arm64 was supported since 3.4.0.
+            # Darwin arm64 was supported since 4.1.0.
+            supported_machines.append("arm64")
+    elif operating_system in ("darwin", "linux"):
+        # This is needed to run bazelisk_test.sh on Linux and Darwin arm64 machines, which are
+        # becoming more and more popular.
+        # It works becasue all recent commits of Bazel supports arm64 on Darwin and Linux.
```
fweikert, line 227: typo

fweikert, line 227: support

bazelisk.py
```diff
@@ -206,16 +206,36 @@ def determine_executable_filename_suffix():
 
 
 def determine_bazel_filename(version):
+    operating_system = get_operating_system()
```
fweikert, line 209: please move this block into its own function

fweikert (changes requested):

linzhp:
@fweikert addressed your comments. All tests passed.

PR #273, opened 2021-10-07
Consider wrapper when only user- or group-executable

EricMountain:
The bazel wrapper might not be executable by `other`, only
`user` or `group`, depending for instance on the umask of the
user.

Allow for this by checking the user and group executable bits
too. If the file is not executable by the current user, the
run will fail with permission denied.


ruddstevens:
Any update on this PR? This would be a great addition. We are running into this with hosts using restrictive umask settings.

EricMountain:
Seems like a similar fix has been merged in the interim, so closing this PR.

PR #275, opened 2021-10-13
Clarify instructions about wrapper

alexeagle:
Some users directly install a Bazel binary into their PATH, for them it doesn't work


philwo (approved):
Thanks!

PR #280, opened 2021-11-19
Add support for patch versions of Bazel (e.g. `1.2.3-patch4`)

UebelAndre:
I have some patches I maintain in a fork of Bazel and generate binaries with `-patch` suffixes followed by a number (e.g. `4.2.1-patch1`) and would like to start using Bazelisk to download those binaries. This mostly works already except that Bazelisk does not accept the `-patch` patch versions. The change is pretty small so thought I'd open a pull requests to learn whether or not this patching pattern was a good idea or if maybe there's a better way. I also understand not everyone may be doing this so am open to other ideas and approaches for using patched Bazel versions.

### DIFF
--- a/versions/versions.go
+++ b/versions/versions.go
@@ -18,6 +18,7 @@ const (
 
 var (
 	releasePattern       = regexp.MustCompile(`^(\d+)\.(x|\d+\.\d+)$`)
+	patchPattern         = regexp.MustCompile(`^(\d+\.\d+\.\d+)-([\w\d]+)$`)
 	candidatePattern     = regexp.MustCompile(`^(\d+\.\d+\.\d+)rc(\d+)$`)
 	rollingPattern       = regexp.MustCompile(`^\d+\.0\.0-pre\.\d{8}(\.\d+){1,2}$`)
 	latestReleasePattern = regexp.MustCompile(`^latest(?:-(?P<offset>\d+))?$`)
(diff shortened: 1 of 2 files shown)


philwo (approved):
LGTM from my side, Florian what do you think?

UebelAndre:
@fweikert Didn't realize there was a merge conflict here. That's been resolved and this PR is ready for a review if you can find the time 🙏

UebelAndre:
@fweikert @philwo Another friendly ping 😅

fweikert:
Sorry, I've been OOO for two weeks. Will take a look later today

versions/versions.go
```diff
@@ -18,6 +18,7 @@ const (
 
 var (
 	releasePattern       = regexp.MustCompile(`^(\d+)\.(x|\d+\.\d+)$`)
+	patchPattern         = regexp.MustCompile(`^(\d+\.\d+\.\d+)-([\w\d]+)$`)
```
fweikert, line 21: I think we should start with a more restrictive regex and then maybe loosen it later. what about `...-patch\d+`
?

fweikert (approved):

versions/versions.go
UebelAndre: Restricting it to just `-patch` would not satisfy my needs unfortunately 😞 Do you feel strongly that it should be restrictive?

fweikert: No, it's fine as-is.

UebelAndre:
Thank you so much! Do you know when the next release will be?

fweikert:
We can cut a new one this week.

fweikert:
Update: I released v1.11.0 yesterday, including this change.

UebelAndre:
Thank you so much!

PR #289, opened 2022-01-11
Teach custom fallback version behavior

bb010g:
This allows for smoother migrations from globally pinned Bazel to project-local pinned Bazel using Bazelisk.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,11 @@ It uses a simple algorithm:
 - If the environment variable `USE_BAZEL_VERSION` is set, it will use the version specified in the value.
 - Otherwise, if a `.bazeliskrc` file exists in the workspace root and contains the `USE_BAZEL_VERSION` variable, this version will be used.
 - Otherwise, if a `.bazelversion` file exists in the current directory or recursively any parent directory, it will read the file and use the version specified in it.
-- Otherwise it will use the official latest Bazel release.
+- Otherwise, if the environment variable `USE_BAZEL_FALLBACK_VERSION` is set to one of the following formats:
+  - If set to a value starting with `error:`, it will report an error and version detection will fail.
+  - If set to a value starting with `warn:`, it will report a warning and use the version specified after the prefix.
+  - If set to a value starting with `silent:`, it will use the version specified after the prefix.
+- Otherwise, it will use the official latest Bazel release.
 
 A version can optionally be prefixed with a fork name.
 The fork and version should be separated by slash: `<FORK>/<VERSION>`.
--- a/core/core.go
+++ b/core/core.go
@@ -250,19 +250,30 @@ func findWorkspaceRoot(root string) string {
 	return findWorkspaceRoot(parentDirectory)
 }
 
+// TODO(go 1.18): remove backport of strings.Cut
+func cutString(s, sep string) (before, after string, found bool) {
+	if i := strings.Index(s, sep); i >= 0 {
+		return s[:i], s[i+len(sep):], true
+	}
+	return s, "", false
+}
+
 func getBazelVersion() (string, error) {
 	// Check in this order:
 	// - env var "USE_BAZEL_VERSION" is set to a specific version.
+	// - workspace_root/.bazeliskrc exists -> read contents, in contents:
+	//   var "USE_BAZEL_VERSION" is set to a specific version.
 	// - env var "USE_NIGHTLY_BAZEL" or "USE_BAZEL_NIGHTLY" is set -> latest
 	//   nightly. (TODO)
 	// - env var "USE_CANARY_BAZEL" or "USE_BAZEL_CANARY" is set -> latest
 	//   rc. (TODO)
 	// - the file workspace_root/tools/bazel exists -> that version. (TODO)
-	// - workspace_root/.bazeliskrc exists and contains a 'USE_BAZEL_VERSION'
-	//   variable -> read contents, that version.
 	// - workspace_root/.bazelversion exists -> read contents, that version.
 	// - workspace_root/WORKSPACE contains a version -> that version. (TODO)
-	// - fallback: latest release
+	// - env var "USE_BAZEL_FALLBACK_VERSION" is set to a fallback version format.
+	// - workspace_root/.bazeliskrc exists -> read contents, in contents:
+	//   var "USE_BAZEL_FALLBACK_VERSION" is set to a fallback version format.
+	// - fallback version format "silent:latest"
 	bazelVersion := GetEnvOrConfig("USE_BAZEL_VERSION")
 	if len(bazelVersion) != 0 {
 		return bazelVersion, nil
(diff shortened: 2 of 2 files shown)


sluongng:
This should come with some documentation in README

bb010g:
> This should come with some documentation in README

Done.

README.md
```diff
@@ -42,7 +42,11 @@ It uses a simple algorithm:
 - If the environment variable `USE_BAZEL_VERSION` is set, it will use the version specified in the value.
 - Otherwise, if a `.bazeliskrc` file exists in the workspace root and contains the `USE_BAZEL_VERSION` variable, this version will be used.
 - Otherwise, if a `.bazelversion` file exists in the current directory or recursively any parent directory, it will read the file and use the version specified in it.
-- Otherwise it will use the official latest Bazel release.
+- Otherwise, if the environment variable `USE_BAZEL_FALLBACK_VERSION` is set to one of the following formats:
```
sluongng, line 45: nit: Worth to note that this is not limited to Environment Variable, but also configurable through `.bazeliskrc` as well as you used `GetEnvOrConfig()`

core/core.go
```diff
@@ -262,6 +262,7 @@ func getBazelVersion() (string, error) {
 	//   variable -> read contents, that version.
 	// - workspace_root/.bazelversion exists -> read contents, that version.
 	// - workspace_root/WORKSPACE contains a version -> that version. (TODO)
+	// - env var "USE_BAZEL_FALLBACK_VERSION" is set to a version.
```
sluongng, line 265: same issue here regarding missing logic about `.bazeliskrc`

core/core.go
```diff
@@ -296,7 +297,22 @@ func getBazelVersion() (string, error) {
...
+		return "", fmt.Errorf("not allowed to use fallback version %q", bazelFallbackVersion)
+	} else if strings.HasPrefix(bazelFallbackVersion, "warn:") {
+		bazelFallbackVersion = bazelFallbackVersion[5:]
+		log.Printf("Warning: used fallback version %q\n", bazelFallbackVersion)
+	} else if strings.HasPrefix(bazelFallbackVersion, "silent:") {
+		bazelFallbackVersion = bazelFallbackVersion[7:]
+	} else {
+		return "", fmt.Errorf("invalid fallback version format %q", bazelFallbackVersion)
+	}
+	return bazelFallbackVersion, nil
```
sluongng, line 315: nit: generally in Golang we prefer to avoid using `else / else if` where possible.

Here you could rewrite each of the `else / else if` clause with an early `return` which should make the code a lot easier to read.

Also the reassignment of `bazelFallbackVersion` could generally be avoided to improve readability as well.

Here is how I would rewrite this block to make it easier to read.
```
	bazelFallbackVersion := GetEnvOrConfig("USE_BAZEL_FALLBACK_VERSION")
	if len(bazelFallbackVersion) != 0 {
		if strings.HasPrefix(bazelFallbackVersion, "error:") {
			return "", fmt.Errorf("not allowed to use fallback version %q", bazelFallbackVersion[6:])
		}
		if strings.HasPrefix(bazelFallbackVersion, "warn:") {
			v := bazelFallbackVersion[5:]
			log.Printf("Warning: used fallback version %q\n", v)

			return v, nil
		}
		if strings.HasPrefix(bazelFallbackVersion, "silent:") {
			return bazelFallbackVersion[7:], nil
		}

		return "", fmt.Errorf("invalid fallback version format %q", bazelFallbackVersion)
	}

	return "latest", nil
```

sluongng (changes requested):
A few more nits

core/core.go
sluongng: can probably do `s/bazelFallbackVersion/fallbackVersion/` as well

bb010g:
Bump.

philwo:
@bb010g It seems like an interesting feature, but it's not obvious to me how to use it - could you give an example how to use this in practice?

bb010g:
@philwo My organization is currently using Bazel 3.3.0 via a pinned, non-Bazelisk installation in CI. That CI setup is used for multiple repositories, so a coordinated upgrade is difficult. We would like to move to Bazelisk-based pinning for these repositories and just install Bazelisk to CI. Without this patch, moving to Bazelisk means that every consumer of this CI installation has to set up `.bazelversion` files in every Bazel workspace before the move to avoid breaking their builds; this can be difficult with our current CI. With this patch, moving to Bazelisk means that CI can `export USE_BAZEL_FALLBACK_VERSION="warn:3.3.0"`, everyone's builds will still work as before by default, CI now tells you when you haven't set up a `.bazelversion` file (or alternatives) for a workspace, and teams can pin version 3.3.0 or whatever versions works for their projects in a decentralized manner.

In the future, after teams aren't seeing warnings about Bazel fallback versions in CI, CI can upgrade (for some or all runs) to `export USE_BAZEL_FALLBACK_VERSION="error:"` (hopefully without breaking any builds), ensuring that it's easy to figure out what version of Bazel any workspace uses.

bb010g:
Additionally, on my developer setup, I can install Bazelisk and `export USE_BAZEL_FALLBACK_VERSION="warn:latest"` to keep Bazelisk handy when I need it but also avoid not pinning the Bazel version for a project workspace.

bb010g:
@sluongng Re-requesting review. (GitHub's button isn't doing anything.)

philwo:
Thank you for the detailed explanation! I think that's pretty cool. @fweikert WDYT?

philwo (approved):

bb010g:
Bump.

core/core.go
```diff
@@ -296,7 +297,22 @@ func getBazelVersion() (string, error) {
 		}
 	}
 
-	return "latest", nil
+	fallbackVersion := os.Getenv("USE_BAZEL_FALLBACK_VERSION")
+	if len(fallbackVersion) == 0 {
+		fallbackVersion = "silent:latest"
+	}
+	if strings.HasPrefix(fallbackVersion, "error:") {
```
fweikert, line 304: it would be easier to split fallbackVersion on the colon and then (if there are two parts) look at part 0 and 1

core/core.go
```diff
@@ -296,7 +297,22 @@ func getBazelVersion() (string, error) {
 		}
 	}
 
-	return "latest", nil
+	fallbackVersion := os.Getenv("USE_BAZEL_FALLBACK_VERSION")
```
fweikert, line 300: is there any downside to accepting USE_BAZEL_FALLBACK_VERSION from the .bazeliskrc file, too?

fweikert (changes requested):

core/core.go
bb010g: That sounds fine. I didn't deal with `.bazeliskrc` interpretation here because the current code doesn't, and I think that should be implemented in a separate PR.

bb010g: I'll note that this means we read the `.bazeliskrc` file up to twice in this function, which is pretty alright for now but it should probably only be read once if `USE_BAZEL_NIGHTLY` or `USE_BAZEL_CANARY` get implemented.

core/core.go
```diff
@@ -250,19 +250,30 @@ func findWorkspaceRoot(root string) string {
 	return findWorkspaceRoot(parentDirectory)
 }
 
+// TODO(go 1.18): remove backport of strings.Cut
+func strings_Cut(s, sep string) (before, after string, found bool) {
```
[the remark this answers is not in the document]
bb010g: `Cut` is simpler.

core/core.go
```diff
@@ -296,7 +307,25 @@ func getBazelVersion() (string, error) {
 		}
 	}
 
-	return "latest", nil
+	fallbackVersionFormat := GetEnvOrConfig("USE_BAZEL_FALLBACK_VERSION")
+	fallbackVersionMode, fallbackVersion, hasFallbackVersionMode := strings_Cut(fallbackVersionFormat, ":")
+	if !hasFallbackVersionMode {
+		fallbackVersionMode, fallbackVersion, hasFallbackVersionMode = "silent", fallbackVersionMode, true
```
fweikert, line 313: no need to assign hasFallbackVersionMode here

core/core.go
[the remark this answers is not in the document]
fweikert: I'm still in favor of Split since it would avoid the custom function for now. If you want to keep this function, then please give it a name that is consistent with the others (otherwise our linter will complain), e.g. cutString

fweikert (approved):

bb010g:
Bump?

PR #292, opened 2022-01-20
Parse .netrc for HTTP Basic authentication credentials when BAZELISK_BASE_URL points to protected resource

### LINKED ISSUE #291: Support using netrc file for bazel downloads in go version
According to my testing done on Windows 10 with Bazelisk 1.11.0, Bazelisk doesn't use the .netrc file for downloading bazel in the go version, but supports it in the python version.

I'd like to request if it's possible to add this support to the go version.

## Testing

For testing I created a small webserver that just prints all headers passed in.
Then I set the following env vars:

- $env:BAZELISK_BASE_URL="http://localhost:3000"
- $env:USE_BAZEL_VERSION="4.1.0"

And fill my C:\Users\<username>\.netrc with the following (also doesn't work when I use a machine specific config):

```netrc
default
login username
password mypassword
```

If I then run any bazelisk command, I can see a request to my test server, but only these headers get set:

| Header | Value |
|:- |:- |
| host | localhost:3000 |
| user-agent | Bazelisk/v1.11.0 |
| accept-encoding | gzip |

## Existing in python version

Support for .netrc was already requested in #247 and implemented for the python version in #269.
It would be awesome if the same feature existed in the go version.

Warchant:
This PR adds support for using .netrc file for getting credentials for Basic HTTP authentication to download bazel from protected resource.

Same as https://github.com/bazelbuild/bazelisk/pull/269 but for Go code.

Fixes #291 
Fixes #269 

Tested with JFrog Artifactory.

### DIFF
--- a/httputil/httputil.go
+++ b/httputil/httputil.go
@@ -139,7 +139,7 @@ func parseRetryHeader(value string) (time.Duration, error) {
 }
 
 // Attempt to open ~/.netrc file and find login/password for given machine (Host).
-func TryFindNetrcFileCreds(machine string) (string, error) {
+func tryFindNetrcFileCreds(machine string) (string, error) {
 	dir, err := homedir.Dir()
 	if err != nil {
 		return "", err
--- a/httputil/httputil.go
+++ b/httputil/httputil.go
@@ -156,9 +156,11 @@ func tryFindNetrcFileCreds(host string) (string, error) {
 	if m == nil {
 		// if host is not found, we should proceed without providing any Authorization header,
 		// because remote host may not have auth at all.
-		log.Printf("Skipping basic authentication for %s because no credentials found in %s", machine, file)
+		log.Printf("Skipping basic authentication for %s because no credentials found in %s", host, file)
 		return "", fmt.Errorf("could not find creds for %s in netrc %s", host, file)
 	}
+
+	log.Printf("Using basic authentication credentials for host %s from %s", host, file)
 
 	login := m.Get("login")
 	pwd := m.Get("password")
--- a/httputil/httputil.go
+++ b/httputil/httputil.go
@@ -2,24 +2,29 @@
 package httputil
 
 import (
+	b64 "encoding/base64"
 	"fmt"
 	"io"
 	"io/ioutil"
 	"log"
 	"math/rand"
 	"net/http"
+	"net/url"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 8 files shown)


Warchant:
How to add dependency to https://github.com/jdxcode/netrc if it doesn't have tags? >.<

minmichail:
@fweikert sorry for pushing that already, but we will need that feature soon and we would like to use an official release in case review pass

httputil/httputil.go
```diff
@@ -153,8 +183,22 @@ func DownloadBinary(originURL, destDir, destFile string) (string, error) {
...
+			panic(err)
+		}
+
 		log.Printf("Downloading %s...", originURL)
-		resp, err := get(originURL, "")
+
+		var auth string = ""
+		t, err := tryFindNetrcFileCreds(u.Host)
+		if err == nil {
+			// successfully parsed netrc for given host
```
fweikert, line 197: I'd fail here if the .netrc file exists, but could not be read (for whatever reasons).

fweikert: It would also be good to add a log statement regarding whether a .netrc file was read or not

httputil/httputil.go
```diff
@@ -153,8 +183,22 @@ func DownloadBinary(originURL, destDir, destFile string) (string, error) {
 			}
 		}()
 
+		u, err := url.Parse(originURL)
+		if err != nil {
+			// originURL supposed to be valid
+			panic(err)
```
fweikert, line 189: Please return an error instead

httputil/httputil.go
```diff
@@ -133,6 +138,31 @@ func parseRetryHeader(value string) (time.Duration, error) {
 	return time.Until(t), nil
 }
 
+// Attempt to open ~/.netrc file and find login/password for given machine (Host).
```
fweikert, line 141: Please start the comment with the function name, otherwise our linter will fail.

httputil/httputil.go
Warchant: >I'd fail here if the .netrc file exists, but could not be read (for whatever reasons).

.netrc may exist, but may not contain given "Host". In this case we should proceed with normal auth, without a failure, because remote may not have auth at all.

Warchant:
Ping @fweikert

fweikert (approved):

PR #295, opened 2022-01-24
Verifying the hash of downloaded Bazel binary

linzhp:
In some areas with unstable or slow internet, the downloaded Bazel may be corrupted. We should verify the hash of the downloaded binary, delete the binary and prompt users to try again if it is corrupted. For backward compatibility for existing mirrors that don't have sha256 file, we will skip the verification if the ".sha256" file doesn't exist.

### DIFF
--- a/bazelisk.py
+++ b/bazelisk.py
@@ -329,7 +329,7 @@ def download_bazel_into_directory(version, is_commit, directory):
         )
         # Exiting with a special exit code not used by Bazel, so the calling process may retry based on that.
         # https://docs.bazel.build/versions/0.21.0/guide.html#what-exit-code-will-i-get
-        sys.exit(45)
+        sys.exit(22)
     return destination_path
 
 
--- a/bazelisk.py
+++ b/bazelisk.py
@@ -18,6 +18,7 @@ limitations under the License.
 import base64
 from contextlib import closing
 from distutils.version import LooseVersion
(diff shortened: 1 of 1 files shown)


linzhp:
@philwo Would you mind also taking a look at this?

linzhp:
@fweikert Would you mind taking a look at this too?

bazelisk.py
```diff
@@ -268,30 +272,59 @@ def download_bazel_into_directory(version, is_commit, directory):
...
+    actual_hash = sha256_hash.hexdigest()
+    if actual_hash != expected_hash:
+        os.remove(destination_path)
+        os.remove(sha256_path)
+        print(
+            "The downloaded Bazel binary is corrupted. Expected SHA-256 {}, got {}. Please try again.".format(
+                expected_hash, actual_hash
+            )
+        )
+        # Exiting with a special exit code not used by Bazel, so the calling process may retry based on that.
```
fweikert, line 305: Current Bazel versions use 45: https://docs.bazel.build/versions/5.0.0/guide.html#what-exit-code-will-i-get

fweikert (changes requested):

bazelisk.py
linzhp: Changed it to 22

fweikert (approved):

PR #309, opened 2022-03-26
build: bump bazel to 5.1.1

chenrui333:


.bazelversion
```diff
@@ -1 +1 @@
-5.0.0
+5.1.1
```
aiuto, line 1: Why not use `5.x` so we track the latest released version.

On the other hand, this is not my call. I don't do the bazelisk releases. I just think that manually tracking this is brittle.

chenrui333: Yeah, that is pain of maintaining the supply chain, but it would be always good to pin the version.

On the other hand, I dont think bazelversion does support `5.x` syntax, correct me if I am wrong.

philwo (approved):

PR #315, opened 2022-04-14
fix #267, BAZELISK_NOJDK support

### LINKED ISSUE #267: bazel nojdk version

honwen:
fix `BAZELISK_NOJDK` support

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,8 @@ As mentioned in the previous section, the `<FORK>/<VERSION>` version format allo
 If you want to create a fork with your own releases, you have to follow the naming conventions that we use in `bazelbuild/bazel` for the binary file names.
 The URL format looks like `https://github.com/<FORK>/bazel/releases/download/<VERSION>/<FILENAME>`.
 
+If the environment variable `BAZELISK_NOJDK` is set and `!= 0`, it will use the `nojdk` version of bazel.
+
 You can also override the URL by setting the environment variable `$BAZELISK_BASE_URL`. Bazelisk will then append `/<VERSION>/<FILENAME>` to the base URL instead of using the official release server. Bazelisk will read file [`~/.netrc`](https://everything.curl.dev/usingcurl/netrc) for credentials for Basic authentication.
 
 ## Ensuring that your developers use Bazelisk rather than Bazel
--- a/platforms/platforms.go
+++ b/platforms/platforms.go
@@ -3,9 +3,11 @@ package platforms
 
 import (
 	"fmt"
-	semver "github.com/hashicorp/go-version"
 	"log"
+	"os"
 	"runtime"
+
+	semver "github.com/hashicorp/go-version"
 )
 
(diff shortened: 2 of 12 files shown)


platforms/platforms.go
```diff
@@ -54,7 +56,11 @@ func DetermineBazelFilename(version string, includeSuffix bool) (string, error)
 		filenameSuffix = DetermineExecutableFilenameSuffix()
 	}
 
-	return fmt.Sprintf("bazel-%s-%s-%s%s", version, osName, machineName, filenameSuffix), nil
+	bazelFlavor := "bazel"
+	if val, ok := os.LookupEnv("BAZELISK_NOJDK"); ok && (val != "0") {
```
ruddstevens, line 60: This would be great to fall back to read it from the config as well with getEnvOrConfig : https://github.com/bazelbuild/bazelisk/blob/70694fe248380c6736a2947a52bcc9344dbfb365/core/core.go#L182

honwen: done. but package `core` has to be splited to break import loop

README.md
```diff
@@ -77,6 +77,8 @@ As mentioned in the previous section, the `<FORK>/<VERSION>` version format allo
 If you want to create a fork with your own releases, you have to follow the naming conventions that we use in `bazelbuild/bazel` for the binary file names.
 The URL format looks like `https://github.com/<FORK>/bazel/releases/download/<VERSION>/<FILENAME>`.
 
+If the environment variable `BAZELISK_NOJDK` is set and `!= 0`, it will use the `nojdk` version of bazel.
```
sluongng, line 80:
```suggestion
If `BAZELISK_NOJDK` is set in environment variable or `.bazeliskrc` to a value that is not `"0"`, bazelisk will use the `nojdk` version of bazel.
```

platforms/platforms.go
```diff
@@ -59,7 +62,11 @@ func DetermineBazelFilename(version string, includeSuffix bool) (string, error)
 		filenameSuffix = DetermineExecutableFilenameSuffix()
 	}
 
-	return fmt.Sprintf("bazel-%s-%s-%s%s", version, osName, machineName, filenameSuffix), nil
+	bazelFlavor := "bazel"
+	if configs.GetEnvOrConfig("BAZELISK_NOJDK") != "0" && configs.GetEnvOrConfig("BAZELISK_NOJDK") != "" {
```
sluongng, line 66: why don't we just set the convention to that `BAZELISK_NOJDK=1` is accepted, and everything else is rejected?

That would make this a lot easier to use/manage vs having to read the code to find these 2 magic values to avoid.

If we want to stay consistent with the python implementation then we would want GetEnvOrConfig here to accept a default return value in case both env and config were not set?

sluongng (commented):
I would love to get this merged early as I want to start doing some experimentation with nojdk bazel. Thanks!

PR #322, opened 2022-06-06
Stop forwarding SIGINT to child processes

### LINKED ISSUE #307: Go bazelisk causes double-SIGINT in "bazel run"
With this `py_binary()`:
```python
import time
import signal

running = True

def _sigint_handler_debounced(signal_number, frame):
    global running
    if not running:
        print("Ignoring superfluous SIGINT")
        return

    print("Stopping the program")
    running = False


signal.signal(signal.SIGINT, _sigint_handler_debounced)

while running:
    time.sleep(1)

print("Almost there. Waiting for potential superfluous SIGINTs")
time.sleep(1)
```
And then when I run it and hit CTRL-C once:
```console
$ bazel run //common/python/double_sigint:foo
^CStopping the program
Ignoring superfluous SIGINT
Almost there. Waiting for potential superfluous SIGINTs
```
This appears to be caused by the signal forwarding in the Go version:
https://github.com/bazelbuild/bazelisk/blob/15abf0e20b43ca0e51e54d43fce916a7f227dc1f/core/core.go#L438-L442

The Python version notably does not have this behaviour:
https://github.com/bazelbuild/bazelisk/blob/81f38ed1a2cb20748a6037d0f81fe943295e1112/bazelisk.py#L445-L452

I'd be tempted to say that bazelisk should only forward SIGTERM and not forward SIGINT.
I.e. hitting CTRL-C on an interactive program should only generate a single SIGINT in the running program. But sending a SIGTERM to bazelisk (via child PID after fork/exec) should also trigger a SIGTERM in the running process.

What do the maintainers here think?

gibfahn:
---

#### Commits _(oldest to newest)_

22c287a Stop forwarding SIGINT to child processes

Since we handle os.Interrupt but no longer forward it to child
processes, this commit also removes platform-specific handling for
Windows, preferring to use child.Process.Kill() in all cases.

Fixes #307.

<br/>

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -14,7 +14,6 @@ import (
 	"os/signal"
 	"path/filepath"
 	"regexp"
-	"runtime"
 	"sort"
 	"strings"
 	"sync"
@@ -438,9 +437,9 @@ func runBazel(bazel string, args []string, out io.Writer) (int, error) {
 	signal.Notify(c, os.Interrupt, syscall.SIGTERM)
 	go func() {
 		s := <-c
-		if runtime.GOOS != "windows" {
-			cmd.Process.Signal(s)
-		} else {
+
+		// Only forward SIGTERM to our child process.
+		if s != os.Interrupt {
 			cmd.Process.Kill()
 		}
 	}()


gibfahn:
@fweikert any chance you could take a look? Hopefully a pretty straightforward change.

fweikert:
Thank you! Looks good to me.

PR #343, opened 2022-08-01
Update Bazel and all dependencies to latest

philwo:
We were lagging behind quite a bit, but this brings everything back to latest versions.

I manually verified that the change does not negatively affect the size of the generated binaries or the structure of the NPM package.


fweikert (approved):

fweikert:
Nice, thank you!

I triggered a rerun of the failing Windows test.

philwo:
Huh... that failure on Windows is strange:

> compilepkg: could not create directory for _empty.go: mkdir C:\b\2hhx3mgx\execroot\__main__\bazel-out\x64_windows-fastbuild\bin\platforms\github_com_bazelbuild_bazelisk_platforms: Cannot create a file when that file already exists.

I don't immediately have an idea what might cause it, but I can try a few things in this PR later or in the next days to fix this.

meteorcloudy:
The error seemed to be through at https://github.com/bazelbuild/rules_go/blob/ed56d7e5b6a7bf833169372742c1ff72f501846a/go/tools/builders/compilepkg.go#L206-L208, if I try to build `//platforms:platforms_test` alone, it doesn't throw the error. I think there is some conflict between actions.

fweikert:
Closing in favor of https://github.com/bazelbuild/bazelisk/pull/403

PR #348, opened 2022-08-17
update rules_nodejs version

Wyverald:


BUILD
```diff
@@ -193,6 +193,7 @@ genrule(
 
 pkg_npm(
     name = "npm_package",
+    package_name = "@bazel/bazelisk",
```
fweikert, line 196: is this needed?

fweikert (approved):

BUILD
Wyverald: yes, the new rules_nodejs version complains if this isn't added.

Wyverald: See https://buildkite.com/bazel/bazelisk/builds/1707

PR #349, opened 2022-08-29
Support BAZELISK_INCOMPATIBLE_FLAGS for explicitly setting incompatible flags to be tested

meteorcloudy:
This makes testing incompatible flags more flexible, both on CI or on users' local machines.

Working towards https://github.com/bazelbuild/continuous-integration/issues/1410

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,6 +113,8 @@ bazelisk --strict build //...
 If the code fails with `--strict`, the flag `--migrate` will run Bazel with each one of the flag separately, and print a report at the end.
 This will show you which flags can safely enabled, and which flags require a migration.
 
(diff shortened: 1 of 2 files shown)


fweikert:
I wonder if we should simply revert https://github.com/bazelbuild/bazelisk/commit/7c8d878a372975279629b5bf70f791d167b803aa and have Bazelisk fetch the incompatible flags from GitHub.

fweikert (approved):

meteorcloudy:
> I wonder if we should simply revert https://github.com/bazelbuild/bazelisk/commit/7c8d878a372975279629b5bf70f791d167b803aa and have Bazelisk fetch the incompatible flags from GitHub.

I guess that could be an additional feature of Bazelisk. I still like to have this env var support because it's more flexible, so the user can test whatever incompatible flags they want to test. For example, on the CI pipeline I can start a new build to test incompatible flag that I'm interested in.

@fweikert Can you help making a new release and deploy to CI! 😃

PR #353, opened 2022-09-02
Ensure that Bazelisk can be built with both Bazel and "go build"

fweikert:


meteorcloudy (approved):

sluongng:
@fweikert Unfortunately this fix is insufficient for another Bazel repository to use `bazelisk` via `go_repository`.

The main problem is the netrc deps will be provided by Gazelle like this:

```
    go_repository(
        name = "com_github_bgentry_go_netrc",
        importpath = "github.com/bgentry/go-netrc",
        sum = "h1:xDfNPAt8lFiC1UJrqV3uuy861HCTo708pDMbjHHdCas=",
        version = "v0.0.0-20140422174119-9fd32a8b3d3d",
    )
```

and thus, there is no prefix trim happening. 
And because BUILD files already exist in Bazelisk, we could not override how import works using a `gazelle:resolve` directive.

Worth to note that compare to Gazelle's `go_repository`, the current approach does not provide any checksum validation on the external dependencies, and it depends on Github's unstable SHA to download the archive. https://blog.bazel.build/2023/02/15/github-archive-checksum.html

Since the netrc library is relatively small, would you consider just... vendoring the parser logic within Bazelisk?

Alternatively, I raised https://github.com/git-lfs/go-netrc/pull/2 to add a go.mod into a more mature fork of the netrc library. Given `git-lfs` reputation, I think switching to it would be a more reliable choice moving forward.

Let me know what you think.

PR #355, opened 2022-09-02
fix go build

chenrui333:
- relates to https://github.com/Homebrew/homebrew-core/pull/109505


fweikert:
Thanks, but I'm closing this PR in favor of https://github.com/bazelbuild/bazelisk/pull/353 since it would break bazel build.

fweikert:
v1.13.1 has the fix :)

PR #361, opened 2022-09-07
Use Linux binaries built on CentOS 7

fweikert:
As a result, Bazelisk should work on Ubuntu 14.04 and 16.04 again. Moreover, CentOS 7 is the most likely candidate for publishing arm64 Bazel binaries in the future.


meteorcloudy (approved):
Thanks!

PR #362, opened 2022-09-09
Pass startup options to bazel shutdown and bazel clean --expunge

meteorcloudy:
If startup options are missing, bazel shutdown/clean commands may end up shutdown the wrong server and clean up wrong caches if flags like --output_user_root is set.

This is needed for fixing Windows permission denied error found at https://buildkite.com/bazel/bazelisk-plus-incompatible-flags/builds/1246#01831fc5-a439-4894-9418-0c4d7591fcfd

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -539,35 +539,34 @@ func insertArgs(baseArgs []string, newArgs []string) []string {
 
 func parseStartupOptions(baseArgs []string) []string {
 	var result []string
-	var BAZEL_COMMANDS = []string {
-		"analyze-profile",
-		"aquery",
-		"build",
-		"canonicalize-flags",
-		"clean",
-		"coverage",
-		"cquery",
-		"dump",
-		"fetch",
-		"help",
-		"info",
-		"license",
-		"mobile-install",
-		"mod",
-		"print_action",
-		"query",
-		"run",
-		"shutdown",
-		"sync",
-		"test",
-		"version",
+	var BAZEL_COMMANDS = map[string]bool{
+		"analyze-profile": true,
+		"aquery": true,
+		"build": true,
+		"canonicalize-flags": true,
+		"clean": true,
+		"coverage": true,
+		"cquery": true,
+		"dump": true,
+		"fetch": true,
+		"help": true,
+		"info": true,
+		"license": true,
+		"mobile-install": true,
+		"mod": true,
+		"print_action": true,
+		"query": true,
+		"run": true,
+		"shutdown": true,
+		"sync": true,
+		"test": true,
+		"version": true,
 	}
 	// Arguments before a Bazel command are startup options.
 	for _, arg := range baseArgs {
-		for _, cmd := range BAZEL_COMMANDS {
-			if arg == cmd {
-				return result
-			}
+		_, prs := BAZEL_COMMANDS[arg]
+		if prs {
+			return result
 		}
 		result = append(result, arg)
 	}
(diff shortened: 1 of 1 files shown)


core/core.go
```diff
@@ -537,14 +537,52 @@ func insertArgs(baseArgs []string, newArgs []string) []string {
 	return result
 }
 
-func shutdownIfNeeded(bazelPath string) {
+func parseStartupOptions(baseArgs []string) []string {
+	var result []string
+	var BAZEL_COMMANDS = []string {
```
fweikert, line 542: Can you turn this into a map for faster lookup?

core/core.go
```diff
@@ -555,14 +593,15 @@ func shutdownIfNeeded(bazelPath string) {
...
-func cleanIfNeeded(bazelPath string) {
+func cleanIfNeeded(bazelPath string, startupOptions []string) {
 	bazeliskClean := GetEnvOrConfig("BAZELISK_CLEAN")
 	if len(bazeliskClean) == 0 {
 		return
 	}
 
-	fmt.Printf("bazel clean --expunge\n")
-	exitCode, err := runBazel(bazelPath, []string{"clean", "--expunge"}, nil)
+	args := append(startupOptions, []string{"clean", "--expunge"}...)
```
fweikert, line 602: append(startupOptions, "clean", "--expunge")

core/core.go
```diff
@@ -575,11 +614,13 @@ func cleanIfNeeded(bazelPath string) {
 
 // migrate will run Bazel with each flag separately and report which ones are failing.
 func migrate(bazelPath string, baseArgs []string, flags []string) {
+	var startupOptions = parseStartupOptions(baseArgs)
```
fweikert, line 617: Do we need this behavior for non-migrate code paths, too?

core/core.go
meteorcloudy: Done!

core/core.go
meteorcloudy: Done!

core/core.go
meteorcloudy: This is only for running `bazel shutdown` and `bazel clean --expunge`, which is only used in migrate code path I think,

core/core.go
```diff
@@ -539,35 +539,34 @@ func insertArgs(baseArgs []string, newArgs []string) []string {
...
+		"test": true,
+		"version": true,
 	}
 	// Arguments before a Bazel command are startup options.
 	for _, arg := range baseArgs {
-		for _, cmd := range BAZEL_COMMANDS {
-			if arg == cmd {
-				return result
-			}
+		_, prs := BAZEL_COMMANDS[arg]
```
fweikert, line 567: What does "prs" stand for? Usually it's "ok"

core/core.go
```diff
@@ -539,35 +539,34 @@ func insertArgs(baseArgs []string, newArgs []string) []string {
...
+		if prs {
```
fweikert, line 568: you can merge the two lines:

if _, prs := BAZEL_COMMANDS[arg]; prs {
 ...
}

fweikert (changes requested):
two nits

core/core.go
meteorcloudy: Done,  thanks!

fweikert (approved):

fweikert:
Presubmit is failing since it's picking exactly the commit whose binary has not been published (due to the gcc platform renaming) -.-

meteorcloudy:
@fweikert Do you have time to create a release to contain this change?

PR #368, opened 2022-09-19
build: bump bazel to 5.3.2

chenrui333:


chenrui333:
actually running into [some issue with 5.3.1 build](https://github.com/Homebrew/homebrew-core/pull/111955)

PR #379, opened 2022-09-24
Eliminate use of deprecated `distutils.version` module

jbms:
Since the recommended replacement, `packaging.version`, is not in the standard library, this commit instead just implements a simple version number comparison without any additional dependencies.

### DIFF
--- a/bazelisk.py
+++ b/bazelisk.py
@@ -17,7 +17,6 @@ limitations under the License.
 
 import base64
 from contextlib import closing
-from distutils.version import LooseVersion
 import hashlib
 import json
 import netrc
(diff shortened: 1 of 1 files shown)


bazelisk.py
```diff
@@ -171,15 +170,18 @@ def read_remote_text_file(url):
...
-    ordered = sorted(
+    return sorted(
         (
-            LooseVersion(release["tag_name"])
+            release["tag_name"]
             for release in get_releases_json(bazelisk_directory)
             if not release["prerelease"]
         ),
+        # This only handles versions with numeric components, but that is fine
+        # since prerelease verisons have been excluded.
```
fweikert, line 180: nit: versions

jbms: Fixed.

fweikert:
Thank you!

fweikert (approved):

PR #390, opened 2022-10-22
Support user `.bazeliskrc` in addition to workspace root `.bazeliskrc`

LINKIWI:
This change introduces support for locating multiple `.bazeliskrc` files, namely, in the workspace root and in the user home (`~/`) directory. Configuration in the former takes precedence over the latter. Semantically, this operates similarly to `.bazelrc`.

This change is backwards-compatible and does not introduce any new behaviors when there is only a single `.bazeliskrc` in the workspace.

The motivation is that I have some configuration (namely, `BAZELISK_BASE_URL`) that I'd like to set globally rather than versioning a `.bazeliskrc` into every Bazel project.

Concretely, here I have:

* Split out `GetEnvOrConfig` into multiple subroutines for locating the relevant RC files, parsing them individually, and loading them into the global configuration map, respecting precedence.
* Restructured existing logic to add `error` returns in the function signature of the above subroutines, and handle this appropriately during the loading phase.
* Added a constant for the `".bazeliskrc"` name.
* Added integration tests to verify the new behavior.

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,7 +129,7 @@ You can control the user agent that Bazelisk sends in all HTTP requests by setti
 
 # .bazeliskrc configuration file
 
-The Go version supports a `.bazeliskrc` file in the root directory of a workspace. This file allows users to set environment variables persistently.
+The Go version supports a `.bazeliskrc` file in the root directory of a workspace and the user home directory. This file allows users to set environment variables persistently.
 
 Example file content:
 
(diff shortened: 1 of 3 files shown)


LINKIWI:
Hm, looks like the integration test I added is failing on Windows. I don't have a Windows machine available to test and not sure how `$HOME` is handled there--if anyone is able to provide some pointers that would be greatly appreciated.

LINKIWI:
Hi team, any chance this change can be looked over? cc @fweikert (tagging since blame suggests you might have the best context around this logic)

fweikert:
Great change, thank you!

fweikert (approved):

PR #395, opened 2022-11-14
Removed unused param in core/core.go downloadBazel function

gregmagolan:
This is just a nit but when I was reading this code yesterday this extra unused param threw me off for a sec


fweikert (approved):

fweikert:
Thanks!

PR #397, opened 2022-11-19
Support aarch64 in `bazelisk.py`.

rpwoodbu:
Linux reports arm64 as aarch64. Correct for this so that Bazel can be fetched when on arm64/aarch64.

### DIFF
--- a/bazelisk.py
+++ b/bazelisk.py
@@ -258,7 +258,7 @@ def normalized_machine_arch_name():
     machine = platform.machine().lower()
     if machine == "amd64":
         machine = "x86_64"
-    if machine == "aarch64":
+    elif machine == "aarch64":
         machine = "arm64"
     return machine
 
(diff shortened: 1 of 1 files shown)


bazelisk.py
```diff
@@ -258,6 +258,8 @@ def normalized_machine_arch_name():
     machine = platform.machine().lower()
     if machine == "amd64":
         machine = "x86_64"
+    if machine == "aarch64":
```
fweikert, line 261: nit: use elif instead

rpwoodbu: Good point. Done.

fweikert (approved):
Thank you!

00z20 (commented):
#423

00z20:
@

> 

> 
Duplicate of #

00z20 (commented):
#397

00z20:
Duplicate of #

Issue #306, opened 2022-03-04
Publish checksum.txt file with each release


bizrad:
It would be nice if there was a checksum.txt file with the contents of the sha256 checksums of each released binary.

e.g. for the 1.11.0 release
```
export BAZELISK_VERSION=1.11.0 && for i in bazelisk-darwin bazelisk-darwin-amd64 bazelisk-darwin-arm64 bazelisk-linux-amd64 bazelisk-linux-arm64 bazelisk-windows-amd64.exe;do curl -LO https://github.com/bazelbuild/bazelisk/releases/download/v${BAZELISK_VERSION}/${i} ; done && sha256sum bazelisk-* >checksums.txt
```

Checksums.txt contents
```
afb0857e4497810dd9695af2b1491c4184debe45ec035a2bb8a4aeb1542166c4  bazelisk-darwin
c725fd574ea723ab25187d63ca31a5c9176d40433af92cd2449d718ee97e76a2  bazelisk-darwin-amd64
1e18c98312d1a03525f704214304be2445478392c8687888d5d37e6a680f31e6  bazelisk-darwin-arm64
231ec5ca8115e94c75a1f4fbada1a062b48822ca04f21f26e4cb1cd8973cd458  bazelisk-linux-amd64
f9119deb1eeb6d730ee8b2e1a14d09cb45638f0447df23144229c5b3b3bc2408  bazelisk-linux-arm64
0eca0630e072434c63614b009bf5b63a7dff489a8676b560ce63e0d91ab731cd  bazelisk-windows-amd64.exe
```

bizrad:
checksums for 0.12.0
```
0f4873534e49db2fc915c127aca86ec5ece0b4ee9042ee5020835be34682661b  bazelisk-darwin
70cf7fe74808d16418d371feb8ccd4e7c44215db03d2c4ffc75b767b765409fb  bazelisk-darwin-amd64
345bb8b900d6b9ef74234e9e9e413aed38fb29ef255e4ae18ac61bf4a6102d84  bazelisk-darwin-arm64
6b0bcb2ea15bca16fffabe6fda75803440375354c085480fe361d2cbf32501db  bazelisk-linux-amd64
29d861ca48df24a3e8dec57fb00508ba664a31921047b2686c38cf9a20d4639f  bazelisk-linux-arm64
f5ace51c34aaf508e32219a07de768b6f6f92f01993800e853a55dc49d3da89c  bazelisk-windows-amd64.exe
```

PR #418, opened 2023-01-23
Generate sha256 message digest for built binaries

ddlees:
Relates to #306 

This work generates files that contain the sha256 message digest for each binary. I'm not sure how you go about populating artifacts to the github release but would be able to do it if someone could point me in the right direction.

### DIFF
--- a/def.bzl
+++ b/def.bzl
@@ -1,0 +1,27 @@
+def _sha256sum_impl(ctx):
+  ctx.actions.run_shell(
+    outputs = [ctx.outputs.sha256],
+    inputs = [ctx.file.src],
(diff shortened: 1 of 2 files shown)


def.bzl
```diff
@@ -0,0 +1,27 @@
+def _sha256sum_impl(ctx):
+  ctx.actions.run_shell(
+    outputs = [ctx.outputs.sha256],
+    inputs = [ctx.file.src],
+    command = "CWD=$PWD && cd `dirname {0}` && sha256sum `basename {0}` > $CWD/{1}".format(ctx.file.src.path, ctx.outputs.sha256.path),
```
fweikert, line 5: We typically build releases on MacOS, which uses a different tool (shasum -a 256)

fweikert:
I think the best approach right now would be to produce a text file with the hashes for each artifact, since we manually upload all artifacts to GitHub. We should probably add a release pipeline to Bazel CI, and might explore something like SLSA in the future.

PR #420, opened 2023-02-02
Normalize machine arch name aarch64 to arm64

yhtang:
Since about 3 days ago, we started [having trouble installing tensorstore on ARM](https://github.com/google/tensorstore/issues/77). Normalizing 'aarch64' to 'arm64' should fix the issue.

### DIFF
--- a/bazelisk.py
+++ b/bazelisk.py
@@ -258,6 +258,8 @@ def normalized_machine_arch_name():
     machine = platform.machine().lower()
     if machine == "amd64":
         machine = "x86_64"
+    if machine == "aarch64":
+        machine = "arm64"
     return machine
 
 


bazelisk.py
```diff
@@ -258,6 +258,8 @@ def normalized_machine_arch_name():
     machine = platform.machine().lower()
     if machine == "amd64":
         machine = "x86_64"
+    if machine == "aarch64":
```
laramiel, line 261: my pending tensorstore patch uses `elif` here

fweikert:
The problem has been fixed by https://github.com/bazelbuild/bazelisk/pull/397

bazelisk.py
yhtang: I updated my commit accordingly. Thanks for the tip. :)

yhtang:
> The problem has been fixed by https://github.com/bazelbuild/bazelisk/pull/397

Oh cool. 🤣

PR #425, opened 2023-02-15
Retry invocation if Bazel exits with code 39

coeuvre:
With https://github.com/bazelbuild/bazel/pull/17358, Bazel will exit with code 39 if remote cache evicts blobs during the build. With https://github.com/bazelbuild/bazel/pull/17462, Bazel is able to continue the build without `bazel clean` or `bazel shutdown`.

This PR makes bazelisk automatically retries the invocation if it sees exit code 39 from Bazel, i.e. "rewinding" on a larger scale.

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -33,6 +33,7 @@ const (
 	wrapperPath    = "./tools/bazel"
 	rcFileName     = ".bazeliskrc"
 	maxDirLength   = 255
+	maxRetryTimes  = 5
 )
 
 var (
@@ -532,7 +533,7 @@ func makeBazelCmd(bazel string, args []string, out io.Writer) *exec.Cmd {
 	return cmd
 }
 
-func runBazel(bazel string, args []string, out io.Writer) (int, error) {
+func runBazelOnce(bazel string, args []string, out io.Writer) (int, error) {
 	cmd := makeBazelCmd(bazel, args, out)
 	err := cmd.Start()
 	if err != nil {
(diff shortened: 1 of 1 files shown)


coeuvre:
cc @tjgq @meisterT

brentleyjones:
This will be a very nice change to go along with the Bazel 6.1 changes.

core/core.go
```diff
@@ -561,6 +562,23 @@ func runBazel(bazel string, args []string, out io.Writer) (int, error) {
...
 
+func isRetriableExitCode(exitCode int) bool {
+	// Bazel exits with 39 when remote cache evicts blobs, in which case, it can recover from error in next invocation.
+	return exitCode == 39
+}
+
+func runBazel(bazel string, args []string, out io.Writer) (int, error) {
+	retries := 0
+	for {
+		exitCode, err := runBazelOnce(bazel, args, out)
```
meteorcloudy, line 573: We should probably print out some message notifying Bazelisk is retrying the bazel invocation?

meteorcloudy (commented):
I wonder if we should guard this behaviour behind a flag? It could be a bit surprising for users that Bazelisk is automatically retrying Bazel invocation.

Also can we add some documentation and ideally a test for this use case?

coeuvre:
The exit code 39 is new in Bazel and is guarded by flag in 6.1.0. I would rather **not** add a new flag here because we certainly want to retry on 39. We can update documentation to mention this behaviour though.

Adding test is tricky. It's hard to make Bazel exits 39 without directly controlling the remote CAS. We have [integration test](https://cs.opensource.google/bazel/bazel/+/master:src/test/java/com/google/devtools/build/lib/remote/BuildWithoutTheBytesIntegrationTest.java;l=470;drc=963640a8d75747c3891911940566cc115eacc340) in Bazel with the help of remote worker. I don't think there is an easy way to mimic it here.

Using a mock also requires code refactor around `runBazel` (e.g. passing a runner interface). If it sounds good, I can go ahead and add a unit test with mock.

meteorcloudy:
Thanks for the context.

> If it sounds good, I can go ahead and add a unit test with mock.

Yes, that sounds good enough!

fweikert:
I wouldn't complain about a test either :)

coeuvre:
Per internal discussion, we probably want to include the retry in Bazel itself. Closing for now.

PR #427, opened 2023-02-16
Implement support for BAZELISK_FORMAT_URL

### LINKED ISSUE #423: Support arbitrary URL layouts
I'm in a situation where I have to download custom Bazel builds, but the only file hosting service I have access to imposes a weird URL layout on me. As a result, I cannot use `BAZELISK_BASE_URL` due to its assumptions on how the repository is laid out.

Could I suggest adding a new `BAZELISK_URL_FMT` variable or similar to generalize the way URLs are computed? It could take something like this:

```BAZELISK_URL_FMT=https://server.example.com/my/weird/<ARCH>/layout/bazel-<RELEASE>-foo/weirder/<OS>```

This new variable should conflict with at least `BAZELISK_BASE_URL` and result in an invalid configuration file error, because the two are incompatible.

Happy to take a stab at implementing this if the idea sounds reasonable.

### PR DESCRIPTION
**Pull request by jmmv**
This new configuration setting provides a format-like string to compute the URL from which to fetch Bazel.  Takes precedence over BAZELISK_BASE_URL as this is a more general concept.

Fixes #423.

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -422,8 +422,14 @@ func downloadBazelIfNecessary(version string, baseDirectory string, repos *Repos
 	}
 
 	var tmpDestPath string
-	if url := GetEnvOrConfig(BaseURLEnv); url != "" {
-		tmpDestPath, err = repos.DownloadFromBaseURL(url, version, destDir, tmpDestFile)
+	baseURL := GetEnvOrConfig(BaseURLEnv)
+	formatURL := GetEnvOrConfig(FormatURLEnv)
+	if baseURL != "" && formatURL != "" {
+		return "", fmt.Errorf("cannot set %s and %s at once", BaseURLEnv, FormatURLEnv)
+	} else if formatURL != "" {
+		tmpDestPath, err = repos.DownloadFromFormatURL(formatURL, version, destDir, tmpDestFile)
+	} else if baseURL != "" {
+		tmpDestPath, err = repos.DownloadFromBaseURL(baseURL, version, destDir, tmpDestFile)
 	} else {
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: core/core.go [hunk 1]
```diff
@@ -422,8 +422,14 @@ func downloadBazelIfNecessary(version string, baseDirectory string, repos *Repos
...
 
 	var tmpDestPath string
-	if url := GetEnvOrConfig(BaseURLEnv); url != "" {
-		tmpDestPath, err = repos.DownloadFromBaseURL(url, version, destDir, tmpDestFile)
+	baseURL := GetEnvOrConfig(BaseURLEnv)
+	formatURL := GetEnvOrConfig(FormatURLEnv)
+	if baseURL != "" && formatURL != "" {
+		return "", fmt.Errorf("cannot set %s and %s at once", BaseURLEnv, FormatURLEnv)
+	} else if formatURL != "" {
+		tmpDestPath, err = repos.DownloadFromFormatURL(formatURL, version, destDir, tmpDestFile)
```
jmmv, line 430: I acknowledge that adding this feels very strange here because we have an input `downloader` function that should encapsulate this. *However* please note that the previous code was already hacking its way to handle the base URL case in the same way, so I just followed existing practice.

I think that this could be generalized to inject these behaviors via `downloader` functions... but I'd rather investigate that separately because there are lots of subtle consequences...

jmmv:
Alright! @fweikert This should now be ready for review again after a rebase + merge conflict resolution. Tests pass. Thanks!

fweikert (approved):
Nice, thank you!

PR #433, opened 2023-03-03
Add fallback to default url

uberzzr:


brentleyjones:
This could lead to non-deterministic versions of bazel being used, right?

uberzzr:
> This could lead to non-deterministic versions of bazel being used, right?

The fallback will use the same version as the version in failed downloading

brentleyjones:
But the binary could be different? Unless I'm misreading and it does the same sha check at both locations, then disregard my comments.

uberzzr:
> But the binary could be different? Unless I'm misreading and it does the same sha check at both locations, then disregard my comments.

Yeah, I see what you mean. I have updated the code, do you think it is correct this time? I assume the bazel file downloaded from the default url will not be corrupted.

bazelisk.py
```diff
@@ -324,16 +338,22 @@ def download_bazel_into_directory(version, is_commit, directory):
...
         os.remove(sha256_path)
         print(
-            "The downloaded Bazel binary is corrupted. Expected SHA-256 {}, got {}. Please try again.".format(
+            "The downloaded Bazel binary is corrupted. Expected SHA-256 {}, got {}. Fallback to default releases.bazel.build url.".format(
                 expected_hash, actual_hash
             )
         )
-        # Exiting with a special exit code not used by Bazel, so the calling process may retry based on that.
-        # https://docs.bazel.build/versions/0.21.0/guide.html#what-exit-code-will-i-get
-        sys.exit(22)
```
yushan26, line 336: Why remove this exit code? It seems like we need it in case the caller retries on this exit code

bazelisk.py
```diff
@@ -324,16 +338,22 @@ def download_bazel_into_directory(version, is_commit, directory):
             sha256_hash.update(byte_block)
     actual_hash = sha256_hash.hexdigest()
     if actual_hash != expected_hash:
-        os.remove(destination_path)
```
yushan26, line 327: Why keep the destination_path folder? Doesn't this mean the binary binary downloaded is corrupted?
We should probably remove the folder and create it again to download the fallback URL

bazelisk.py
```diff
@@ -324,16 +338,22 @@ def download_bazel_into_directory(version, is_commit, directory):
...
+        matched=re.match(r"(\d*\.\d*(?:\.\d*)?)(rc\d+)?", version)
+        if matched:
+            (version, rc) = matched.groups()
+            fallback_url="https://releases.bazel.build/{}/{}/{}".format(
+                version, rc if rc else "release", bazel_filename
+            )
+            download(fallback_url+".sha256", sha256_path)
+            os.remove(destination_path)
+            download(fallback_url,destination_path)
+            os.chmod(destination_path, 0o755)
```
yushan26, line 356: If the download fails for the exit URL, we should probably exit with code 22.

bazelisk.py
```diff
@@ -314,7 +314,21 @@ def download_bazel_into_directory(version, is_commit, directory):
...
+                if "https://releases.bazel.build" not in bazel_url:
+                    matched=re.match(r"(\d*\.\d*(?:\.\d*)?)(rc\d+)?", version)
+                    if matched:
+                        (version, rc) = matched.groups()
+                        fallback_url="https://releases.bazel.build/{}/{}/{}".format(
+                            version, rc if rc else "release", bazel_filename
+                        )
+                        download(fallback_url+".sha256", sha256_path)
+                        os.remove(destination_path)
+                        download(fallback_url,destination_path)
```
yushan26, line 326: What if the fallback URL download fail? Should we have a try catch to raise an exception as well?

yushan26 (changes requested):

bazelisk.py
```diff
@@ -314,7 +314,21 @@ def download_bazel_into_directory(version, is_commit, directory):
...
                 )
-                return destination_path
+                if "https://releases.bazel.build" not in bazel_url:
+                    matched=re.match(r"(\d*\.\d*(?:\.\d*)?)(rc\d+)?", version)
+                    if matched:
+                        (version, rc) = matched.groups()
+                        fallback_url="https://releases.bazel.build/{}/{}/{}".format(
+                            version, rc if rc else "release", bazel_filename
+                        )
+                        download(fallback_url+".sha256", sha256_path)
```
tyler-french, line 324: Let's be consistent and use `format` here for string concatenation
```suggestion
                        download("{}.sha256".format(fallback_url), sha256_path)
```

bazelisk.py
```diff
@@ -314,7 +314,21 @@ def download_bazel_into_directory(version, is_commit, directory):
                 sys.stderr.write(
                     "The Bazel mirror does not have a checksum file; skipping checksum verification."
                 )
-                return destination_path
+                if "https://releases.bazel.build" not in bazel_url:
+                    matched=re.match(r"(\d*\.\d*(?:\.\d*)?)(rc\d+)?", version)
+                    if matched:
```
tyler-french, line 319: We can reduce indentation by negating the logic:

```
if "https://releases.bazel.build" in bazel_url:
    return destination_path
matched = matcher.match( ... )
if !matched:
    return destination_path

(version, rc) = matched.groups()
...

File: bazelisk.py [hunk 7]
```diff
@@ -306,6 +306,8 @@ def download_bazel_into_directory(version, is_commit, directory):
 
     sha256_path = destination_path + ".sha256"
     expected_hash = ""
+    matcher=re.compile(r"(\d*\.\d*(?:\.\d*)?)(rc\d+)?")
```
tyler-french, line 309: Lets move this to the top of the file like line 47:
```
VERSION_PATTERN = re.compile(...
```

uberzzr:
@brentleyjones Can you please review this PR?

brentleyjones:
@fweikert

bazelisk.py
fweikert, line 309: Can you please format the changes? I think running black should work.

uberzzr: Have run black to reformat. Can you please take a look? Thanks!

uberzzr: @fweikert Hi, is there anything else I need to do for this PR? Appreciate a lot if could be merged.

PR #441, opened 2023-03-20
Implement support for BAZELISK_VERIFY_SHA256

jmmv:
The new BAZELISK_VERIFY_SHA256 variable can be set to the expected SHA256 hash of the downloaded Bazel binary.  If set, then the binary is required to match the hash before it is used.

This is important for cases where provenance of the artifact cannot be asserted purely via the HTTPS trust chain (such as what happens in a mutable artifact repository with lax access controls).

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -411,6 +411,9 @@ func downloadBazelIfNecessary(version string, baseDirectory string, repos *Repos
 	}
 
 	destDir := filepath.Join(baseDirectory, pathSegment, "bin")
+	expectedSha256 := strings.ToLower(GetEnvOrConfig("BAZELISK_VERIFY_SHA256"))
+
+	tmpDestFile := "bazel-tmp" + platforms.DetermineExecutableFilenameSuffix()
 	destFile := "bazel" + platforms.DetermineExecutableFilenameSuffix()
 
 	destPath := filepath.Join(destDir, destFile)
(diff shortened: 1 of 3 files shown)


jmmv:
Wohoo. Tests pass now in all platforms and I have rebased this to resolve conflicts. Any thoughts on this @fweikert ?

core/core.go
```diff
@@ -411,18 +411,55 @@ func downloadBazelIfNecessary(version string, baseDirectory string, repos *Repos
...
+		tmpDestPath, err = downloader(destDir, tmpDestFile)
+	}
+	if err != nil {
+		return "", err
+	}
+
+	if len(expectedSha256) > 0 {
+		f, err := os.Open(tmpDestPath)
+		if err != nil {
+			os.Remove(tmpDestPath)
```
fweikert, line 437: What about defer'ing os.Remove(tmpDestPath) and f.Close()?

fweikert (changes requested):
Only one minor thing. Thanks for the great contribution :)

core/core.go
jmmv: Sure, switched to defer for the file deletion.

But I cannot do this for Close. That was what I originally did, but then encountered failures on Windows because keeping the file open interferes with the rename. Added a comment.

fweikert (approved):

jmmv:
@fweikert Thanks for the quick response! Now I'd like to divert your attention to https://github.com/bazelbuild/bazelisk/pull/427 but I first need to fix merge conflicts caused by this PR :)

PR #451, opened 2023-04-28
Support bisecting Bazel to find which Bazel change breaks your build

meteorcloudy:
- Use --bisect=<BASE>..<HEAD> to specify the bisecting range, Bazelisk
  uses the GitHub API to get the list of commits to bisect. You may need
  to set `BAZELISK_GITHUB_TOKEN` to get around GitHub rate limit.
- BAZELISK_SHUTDOWN, BAZELISK_CLEAN can be used to run `bazel shutdown`
  or `bazel clean --expunge` between builds.

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -743,10 +743,6 @@ type CompareResponse struct {
 	MergeBaseCommit Commit `json:"merge_base_commit"`
 }
 
-type Message struct {
-	Content string `json:"message"`
-}
-
 func sendRequest(url string) (*http.Response, error) {
 	client := &http.Client{}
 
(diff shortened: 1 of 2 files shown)


core/core.go
```diff
@@ -140,14 +125,25 @@ func RunBazeliskWithArgsFunc(argsFunc ArgsFunc, repos *Repositories) (int, error
 		if err != nil {
 			return -1, fmt.Errorf("could not get the list of incompatible flags: %v", err)
 		}
-
 		if args[0] == "--migrate" {
 			migrate(bazelPath, args[1:], newFlags)
-		} else {
+		} else if args[0] == "--migrate" {
```
fweikert, line 130: you mean --strict?

fweikert: The else alone is still fine, given the outer if

meteorcloudy: Oh, right, or I should just remove this check.

meteorcloudy: Done

fweikert (approved):

Wyverald:
nice! :D

PR #459, opened 2023-05-10
Retry fetching Bazel versions from GCS URL

meteorcloudy:
Addressing: https://github.com/bazelbuild/continuous-integration/issues/1627

### DIFF
--- a/repositories/gcs.go
+++ b/repositories/gcs.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"log"
 	"strings"
+	"time"
 
 	"github.com/bazelbuild/bazelisk/core"
 	"github.com/bazelbuild/bazelisk/httputil"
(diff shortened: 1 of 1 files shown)


repositories/gcs.go
```diff
@@ -77,7 +77,19 @@ func listDirectoriesInReleaseBucket(prefix string) ([]string, bool, error) {
...
+		var content []byte
+		var err error
+		// Theoretically, this should always work, but we've seen transient
+		// errors on Bazel CI, so we retry a few times to work around this.
+		// https://github.com/bazelbuild/continuous-integration/issues/1627
+		for attempt := 0; attempt < 5; attempt++ {
+			content, _, err = httputil.ReadRemoteFile(url, "")
+			if err == nil {
+				break
+			}
```
fweikert, line 90: can you add a sleep()?

fweikert: not sure if an exponential backoff is required, though

meteorcloudy: Let's do it

fweikert (approved):

PR #465, opened 2023-05-16
Use consistent paths to bazel binaries

illicitonion:
Before this change, changing which mirror you were downloading from would change the $PATH environment variable bazel is run with, even though the bazel binaries being pointed to are identical. This can cause repository rules to invalidate and re-run.

Instead, store downloaded bazels in directories keyed off of the sha256 of the bazel binary itself, and track the metadata of a mirror+version -> sha256.

This avoid spurious rebuilds when only changing the URL bazelisk would use to download bazel.

### DIFF
--- a/bazelisk_test.sh
+++ b/bazelisk_test.sh
@@ -299,6 +299,52 @@ EOF
       (echo "FAIL: Expected to find 'BAZELISK_SKIP_WRAPPER=true' in the output of 'bazelisk version'"; exit 1)
 }
 
+function test_path_is_consistent_regardless_of_base_url() {
+  setup
+
+  echo 6.2.0 > .bazelversion
+
+  cat >WORKSPACE <<EOF
+load("//:print_path.bzl", "print_path")
+
+print_path(name = "print_path")
+
+load("@print_path//:defs.bzl", "noop")
+
+noop()
+EOF
+
+cat >print_path.bzl <<EOF
+def _print_path_impl(rctx):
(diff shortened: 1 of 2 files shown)


illicitonion:
@fweikert could you take a look? Thanks!

fweikert (approved):

fweikert:
Nice, thank you for this PR!

bushuevkonstantun1408 (commented):
Не хотите работать-удалю всё!
Не плохо устроились педики!
Война войной-а обет по распорядку?Ловля моментов закончиться шакадов!

meteorcloudy:
https://github.com/bazelbuild/bazelisk/blob/b7f5c7570df35ce2273afa0c15f4bc5038fb126a/bazelisk_test.sh#L336

Bazelisk postsubmit is now failing with
```
INFO: Starting clean (this may take a while). Consider using --async if the clean takes more than several minutes.
2023/11/10 00:08:53 Downloading https://downloads.sourceforge.net/project/bazel.mirror/6.2.0/bazel-6.2.0-linux-x86_64...
2023/11/10 00:08:53 could not download Bazel: failed to download bazel: failed to download bazel: HTTP GET https://downloads.sourceforge.net/project/bazel.mirror/6.2.0/bazel-6.2.0-linux-x86_64 failed with error 404
```

Any idea how to fix the test?

fweikert:
Looks like sourceforge dropped 6.2.0. We can switch to 6.2.1, but that will lead to the same error in the future.

PR #472, opened 2023-06-01
go_netrc: switch back to gazelle style import

sluongng:
This enables other projects to depend on bazelisk using a gazelle
generated go_repository for com_github_bgentry_go_netrc, instead of
having to use the archive strip_prefix go_repository that Bazelisk
handcrafted.


sluongng:
@fweikert is it possible to get a review from you on this PR? 🙇

fweikert (approved):

PR #486, opened 2023-08-10
Remove rules_go as deps

wisechengyi:
### Problem / Solution

In theory bazelisk should not depend on rules_go, and indeed after a `go mod tidy`, `rules_go` went away. 

I searched around and didn't find too much context why rules_go was added.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,6 @@ module github.com/bazelbuild/bazelisk
 go 1.18
 
 require (
-	github.com/bazelbuild/rules_go v0.39.1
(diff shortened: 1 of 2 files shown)


sluongng:
It was added in https://github.com/bazelbuild/bazelisk/commit/70eecfe6f2b3f0c821f77faae70f231bb3239286 by @philwo. Looks like it's not needed 🤔

wisechengyi:
Hi @fweikert would you kindly see if this is merge-able?

philwo:
Mhm, yeah. Strange. I also can't really say why it would be needed. Removing it LGTM and we'll find out if something breaks ;)

fweikert (approved):

Issue #432, opened 2023-03-03
Flakiness downloading bazel (on windows)


phlax:
Our (Envoy) Windows CI seems to fail fairly frequently downloading the bazel bin:

```console
2023/03/03 08:56:37 Downloading https://releases.bazel.build/6.0.0/release/bazel-6.0.0-windows-x86_64.exe...
2023/03/03 08:56:51 could not download Bazel: HTTP GET https://releases.bazel.build/6.0.0/release/bazel-6.0.0-windows-x86_64.exe failed: Get "[https://releases.bazel.build/6.0.0/release/bazel-6.0.0-windows-x86_64.exe"](https://releases.bazel.build/6.0.0/release/bazel-6.0.0-windows-x86_64.exe%22): net/http: TLS handshake timeout
```

https://dev.azure.com/cncf/envoy/_build/results?buildId=129800&view=logs&j=b840a642-5ff3-5357-2e4b-e06e40b0cffd&t=67965174-5100-5631-9dc0-68b9f0aacb53&l=56

i didnt notice this happening in other jobs - so i believe this is specific to windows - but im not certain of this

either way, im wondering if it  would be possible to add some retry logic/config as i see discussed elsewhere

alyssawilk:
https://dev.azure.com/cncf/envoy/_build/results?buildId=132015&view=results

https://dev.azure.com/cncf/envoy/_build/results?buildId=132024&view=results

celestialorb:
Slapped together a rather rudimentary implementation in #495 as I've been encountering issues while downloading where a basic retry would likely have solved it.

phlax:
cc @akshaysngupta

akshaysngupta:
@celestialorb's PR looks reasonable to me. Can someone from maintainers review this?

PR #495, opened 2023-09-08
Rudimentary Bazel Download Retry Logic

celestialorb:
I get exceptions on downloading Bazel via Bazelisk quite often and I was surprised to see no retry logic implemented for the download, so I quickly slapped together something basic. Hopefully this'll get the ball rolling on a proper solution.

My initial thought was to use something like [`tenacity`](https://pypi.org/project/tenacity/) to wrap the download function, but didn't see an easy way of adding a Python dependency to Bazelisk.

There's also the approach of using the native retry handler with `urllib`, but I'm not sure if that'll work with Python2.


bazelisk.py
```diff
@@ -339,21 +338,30 @@ def download_bazel_into_directory(version, is_commit, directory):
...
+            if "BAZELISK_BASE_URL" in os.environ:
+                parts = urlparse(url)
+                creds = None
+                try:
+                    creds = netrc.netrc().hosts.get(parts.netloc)
+                except Exception:
+                    pass
+                if creds is not None:
+                    auth = base64.b64encode(("%s:%s" % (creds[0], creds[2])).encode("ascii"))
+                    request.add_header("Authorization", "Basic %s" % auth.decode("utf-8"))
```
akshaysngupta, line 356: Can we construct `auth` and `request` once and use them across retries ?

celestialorb: I don't see why not. I've made the changes but I'm hitting issues testing it locally. Don't think it's related to the changes, seems it's just unable to download a specific Bazel checksum file; specifically `https://storage.googleapis.com/bazel-builds/artifacts/ubuntu1404/f3aa4184c0b53864f597ecfb8969938e9a2a8287/bazel.sha256`

I'll go ahead and push my changes and then try reverting them locally to see if the old code in this PR works or if this is indeed an issue accessing that resource.

celestialorb: As a side note I also added configuration of the retry logic via environment variables to this PR. If you'd me to change those variables or how those settings are sourced just let me know.

celestialorb: > I don't see why not. I've made the changes but I'm hitting issues testing it locally. Don't think it's related to the changes, seems it's just unable to download a specific Bazel checksum file; specifically `https://storage.googleapis.com/bazel-builds/artifacts/ubuntu1404/f3aa4184c0b53864f597ecfb8969938e9a2a8287/bazel.sha256`
> 
> I'll go ahead and push my changes and then try reverting them locally to see if the old code in this PR works or if this is indeed an issue accessing that resource.

Figured it out, didn't realize that for the checksum the outer scope was expecting a 404 `HTTPError` to be thrown if the checksum doesn't exist for the specific Bazel version.

celestialorb:
@akshaysngupta @maintainers I went ahead and reexamined this one and saw that one of the tests timed out. I pushed an empty commit to retrigger the job and the test successfully passed after that. Should be ready to be merged in if someone could take a look at it.

celestialorb:
Should at least help to mitigate #432.

celestialorb:
How do we get owner attention on this PR?

alexeagle:
@meisterT this is a priority for some of my users who are running into CI flakiness due to Bazel download failures. Who can look at this?

meisterT:
Let me figure out who can do that. Note that this seems to touch the python version which is lacking behind in features - out of curiosity: why are you not using the golang version?

meteorcloudy:
Sorry for delay, but also curious about @meisterT 's question, are you sure the Python version is what you use?

alexeagle:
In the event we don't hear back from the original author after 2y, I think we would only fix this in the Go implementation. And should probably have a plan to remove the Python implementation or at least mark it obsolete

celestialorb:
> Let me figure out who can do that. Note that this seems to touch the python version which is lacking behind in features - out of curiosity: why are you not using the golang version?

Honestly I'm not sure. We were encountering download failures in our CI with Bazelisk and I probably searched for the given error message which lead me to the Python implementation. What's the difference between the two implementations, and why are there two? Is the Golang implementation newer / preferred? If it helps I checked our version of `bazelisk` in our environment and we're on `v1.17.0`.

meteorcloudy:
@celestialorb Yes, the golang version if preferred and I suspect you are already using it since we don't actually have release for the python version. Please check.

@alexeagle I agree there should be only one version. But currently there are at least three? 
```
bazelisk.go
bazelisk.js
bazelisk.py
```

I have no idea how the js and py version are used. /cc @fweikert

fweikert:
The unofficial policy has been to accept community contributions to the Python version, but we (Bazel team) release new features for the Go version only. The JS version seems to be stale and should be removed.

There's quite a backlog of PRs in general due to a lack of time. I hope this is something we can work through this month.

PR #500, opened 2023-09-21
Make getBazelVersion func public

francine-blanc:
At the moment, `func getBazelVersion()` is scoped to the `core.go` package. I have made a small change to make this function public, so it can be called from other packages.


fweikert (approved):
Also happy to see a test for it in the future ;)

PR #503, opened 2023-10-02
Support `MODULE.bazel` and `REPO.bazel`

### LINKED ISSUE #502: `MODULE.bazel` and `REPO.bazel` are not supported
As of Bazel 6.3.0, `MODULE.bazel` and `REPO.bazel` are also considered workspace boundaries. Bazelisk currently does not search for these, so in situations where either is present but `WORKSPACE.bazel` or `WORKSPACE` is absent, it will fail to find the workspace root and any relative `.bazeliskrc` or `.bazelversion`.

crncnnr:
As of Bazel 6.3.0, `MODULE.bazel` and `REPO.bazel` are also considered workspace boundaries. Bazelisk now searches for these in the same order Bazel does.

Fixes #502

### DIFF
--- a/bazelisk.py
+++ b/bazelisk.py
@@ -95,10 +95,10 @@ def decide_which_bazel_version_to_use():
 def find_workspace_root(root=None):
     if root is None:
         root = os.getcwd()
-    if os.path.exists(os.path.join(root, "WORKSPACE")):
-        return root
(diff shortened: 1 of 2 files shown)


meteorcloudy (approved):
Thanks!

PR #511, opened 2023-10-26
Add progress display during binary downloads.

katre:
Example of progress message:
```
Downloading: 16 MB out of 49 MB (34%)
```

When using stdout this will be overwritten repeatedly.


katre:
I don't understand the test failures: this appears to be fine in Buildkite on the master branch, but when I run the same test at master locally it fails with the same 404 error.

meteorcloudy (approved):

PR #526, opened 2024-01-04
Allow specifying where stdout should get written

illicitonion:
This can be useful when invoking bazelisk programatically.


illicitonion:
@fweikert Any chance of a quick look at this tiny (but slightly ugly!) PR? Thanks!

fweikert (approved):

PR #550, opened 2024-02-02
fix: safely set user agent

sibicramesh:


albertocavalcante:
Hey @sibicramesh, can you please share why this would be beneficial? (asking also cause I assume this func is single-threaded and it's not totally clear to me).

Thanks!

PR #557, opened 2024-02-28
chore(bazel): add MODULE.bazel files for bzlmod

mmorel-35:
Signed-off-by: Matthieu MOREL <matthieu.morel35@gmail.com>

PR #564, opened 2024-03-23
Bump Bazel to 7.1.1

jwnimmer-tri:
Switch to bzlmod exclusively (no WORKSPACE anymore).

Fix bzlmod integration with npm (originally added in #557 but didn't work at all). The 5.x series of rules_js is unmaintained and does not support bzlmod, and the 6.x series only has toolchains (with no packaging rules). Switch to aspect_rules_js which is currently the best source of npm packaging rules.

---

Obviously I haven't tried uploading the new npm package, but looking at the results in `bazel-bin` it seems to be comparable output to what currently happens on `master`.


meteorcloudy (approved):
Awesome, thank you very much!

PR #566, opened 2024-03-28
Exec bazel instead of subprocessing

aaron-skydio:
Instead of running bazel as a subprocess of bazelisk, replace bazelisk with bazel in the same process (in the case when bazelisk is only being used to invoke bazel once and then exit, as opposed to invoking bazel multiple times or post-processing output from bazel).  This eliminates several classes of issues.

A couple concrete examples:

- shells' segfault detection often does not fire on subprocesses, so doing `bazel run //:something_that_segfaults` will not cause bash to display that the process segfaulted if bazelisk subprocesses to bazel, but it does if bazelisk `exec`s bazel.
- terminal emulators that display the name of the running process currently display `bazel`, even once `bazel run` has `exec`d the binary it built, since the running top-level process is still bazelisk; they should display the name of the binary, and they do after this change

`syscall.Exec` is not supported on windows, so this only applies to linux and macos.

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -79,32 +79,83 @@ func MakeDefaultConfig() config.Config {
 }
 
 // RunBazelisk runs the main Bazelisk logic for the given arguments and Bazel repositories.
+//
+// This will run Bazel in a subprocess and return its exit code.
 func RunBazelisk(args []string, repos *Repositories) (int, error) {
 	return RunBazeliskWithArgsFunc(func(_ string) []string { return args }, repos)
 }
 
 // RunBazeliskWithArgsFunc runs the main Bazelisk logic for the given ArgsFunc and Bazel
 // repositories.
+//
+// This will run Bazel in a subprocess and return its exit code.
 func RunBazeliskWithArgsFunc(argsFunc ArgsFunc, repos *Repositories) (int, error) {
-
 	return RunBazeliskWithArgsFuncAndConfig(argsFunc, repos, MakeDefaultConfig())
 }
 
 // RunBazeliskWithArgsFuncAndConfig runs the main Bazelisk logic for the given ArgsFunc and Bazel
 // repositories and config.
+//
+// This will run Bazel in a subprocess and return its exit code.
 func RunBazeliskWithArgsFuncAndConfig(argsFunc ArgsFunc, repos *Repositories, config config.Config) (int, error) {
 	return RunBazeliskWithArgsFuncAndConfigAndOut(argsFunc, repos, config, nil)
 }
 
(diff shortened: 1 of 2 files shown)


tjgq:
@aaron-skydio I like this PR - I think it's a more definitive solution to the signal handling issues I partially fixed in edab5e0f55c68abb26c2603140b1e8951425af93. Mind rebasing onto master so we can get this merged?

aaron-skydio:
Done!

core/core.go
```diff
@@ -90,12 +90,6 @@ func RunBazeliskWithArgsFunc(argsFunc ArgsFunc, repos *Repositories) (int, error
 // RunBazeliskWithArgsFuncAndConfig runs the main Bazelisk logic for the given ArgsFunc and Bazel
 // repositories and config.
 func RunBazeliskWithArgsFuncAndConfig(argsFunc ArgsFunc, repos *Repositories, config config.Config) (int, error) {
-	return RunBazeliskWithArgsFuncAndConfigAndOut(argsFunc, repos, config, nil)
-}
-
-// RunBazeliskWithArgsFuncAndConfigAndOut runs the main Bazelisk logic for the given ArgsFunc and Bazel
-// repositories and config, writing its stdout to the passed writer.
-func RunBazeliskWithArgsFuncAndConfigAndOut(argsFunc ArgsFunc, repos *Repositories, config config.Config, out io.Writer) (int, error) {
```
fweikert, line 98: The removal of this function will break users (https://github.com/bazelbuild/bazelisk/pull/526), so I'd like to keep it.
Also, in the current version the out parameter of runBazel is basically useless since it will always be nil.

aaron-skydio: Ah thanks, I didn't realize bazelisk supported use as a library so I just looked for uses in bazelisk itself.  Added it back.

Are there going to be callers who need to call one of these RunBazelisk functions with `out == nil` and have it return instead of `exec`ing?  The way I've written this right now doesn't support that, and I'm not sure how you'd rather me do that if we need that - duplicate all these public functions with ExecBazelisk copies?  Or add a flag to exec if possible or not?

aaron-skydio: Ok I think if I were a user I'd want to have the behavior of these preserved to return instead of execing, so I've added `Exec` versions of the ones that make sense (without `out`) and left the `Run` ones to still run in a subprocess and return

aaron-skydio:
@fweikert @valco1994 I've updated this again, per https://github.com/bazelbuild/bazelisk/pull/757#issuecomment-4264173809.  Could I get another review @fweikert ?  Are you still interested in merging this?

fweikert (approved):

fweikert:
Thanks!

PR #567, opened 2024-04-02
Allow overriding tools/bazel path with BAZELISK_WRAPPER_DIRECTORY

keith:
In some projects the tools/bazel path isn't practical. This allows users to override this path with the BAZELISK_WRAPPER_DIRECTORY env var / .bazeliskrc config option. This allows overriding the root
directory that the `bazel` wrapper executable should be found in. This was chosen over the file itself
in order to continue supporting recursive `bazel` invocations.

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -33,10 +33,11 @@ import (
 )
 
 const (
-	bazelReal      = "BAZEL_REAL"
-	skipWrapperEnv = "BAZELISK_SKIP_WRAPPER"
-	wrapperPath    = "./tools/bazel"
-	maxDirLength   = 255
+	bazelReal               = "BAZEL_REAL"
+	skipWrapperEnv          = "BAZELISK_SKIP_WRAPPER"
+	defaultWrapperDirectory = "./tools"
+	defaultWrapperName      = "bazel"
+	maxDirLength            = 255
 )
 
 var (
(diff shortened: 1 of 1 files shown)


luispadron (approved):

keith:
@fweikert ptal

jszumski (approved):

brentleyjones:
One downside with this is that scripts that recursively call into `bazel` might break now. They would work before because the tools directory is added to `PATH`. If this allowed specifying the parent directory of the wrapper, but the wrapper had to be called `bazel`, then that use case would continue to work.

keith:
yea i guess that's up to the user to manage if they rely on that. we could alternatively just allow a different directory vs different file name, since I imagine that's the more common thing that's important to override.

luispadron:
I'd be happy with different directory if it helps with the issue Brentley mentioned

brentleyjones:
It wouldn't entirely be in control of the user though, because the scripts I'm talking about are run targets, like rules_xcodeproj's project generation.

So I would prefer to limit it to the directory if possible.

keith:
updated

brentleyjones (approved):

fweikert (approved):

fweikert:
Thank you!

keith:
thanks!

PR #576, opened 2024-05-09
Split lines of .bazelversion and return first nonempty.

API92:
Split lines of .bazelversion and return first nonempty in bazelisk.py to support multiline .bazelversion in tensorflow.


fweikert (approved):

fweikert:
I think if we support multi-line files, then we should also skip comments. However, this can be done in a future PR, so I'm approving now.

Issue #586, opened 2024-06-25
`.bazeliskrc`: Support Variable Expansion and Tilde Notation in Home Directory Paths


albertocavalcante:
Currently if someone sets in `.bazeliskrc` the following home paths:

```txt
BAZELISK_HOME=$HOME/.bazelisk
```

```txt
BAZELISK_HOME=~/.bazelisk
```

It will invoke `os.MkdirAll` with the raw string, therefore creating a `./'$HOME'` and `./'~'` respectively.

I would like to propose supporting both tile and var expansions in the path. Will submit a PR.

PR #587, opened 2024-06-25
`.bazeliskrc`: Add Support for Env Var and Tilde Expansion

albertocavalcante:
This is a proposal to add support for env expansion as `os.MkdirAll` will treat shell variables and `~` as literals.

Fixes https://github.com/bazelbuild/bazelisk/issues/586


albertocavalcante:
Hi @fweikert, could you or someone please review this PR when you have a chance? Thanks!

fweikert (approved):

fweikert:
Ah, I was too quick to merge. Yeah, I would have preferred an early return, too, but didn't want to block the PR on it.

Issue #572, opened 2024-04-28
Bazelisk pulls incorrect Bazel on Windows ARM64


jxy-s:
The x64 version of bazelisk is compatible with a Windows ARM64 machine, it will run under x64 emulation on Windows. However it ends up pulling the x64 version of Bazel which causes unintended downstream behavior.
- toolchains are not resolved as expected
- the host OS is not reflected correctly causing the default compilation to produce an x64 binary instead of an ARM64 one

Bazelisk should pull the ARM64 version of when running on Windows ARM64 instead of the x64 version.

See: https://github.com/bazelbuild/bazel/issues/22164#issuecomment-2081383179

meteorcloudy:
Maybe it's as simple as adding another target for windows arm64 here?
https://github.com/bazelbuild/bazelisk/blob/47f60477721681a117cbf905784ee5220100e68b/BUILD#L145-L153

albertocavalcante:
@meteorcloudy my understanding is the same. I will try to submit a PR for this.

PR #601, opened 2024-08-06
Add Windows ARM64 and Refactored Targets

albertocavalcante:
# Summary

Gazelle targets haven't been updated in a while. Took this opportunity to re-generate them.

Also, the `BUILD` file had many `go_binary` directives even though there are just two flavors: a stripped binary and another with all the symbols. Refactored to dinamically generate those based on the `os`, `arch` we should have for each flavor.

Added `windows-arm64`, tested in a workstation on Azure. Find evidence below.

> [!WARNING]  
> **bazelisk.js**: The NodeJS version installed in the machine must be ARM, otherwise `os.arch()` returns `x64` and it will use the `windows_amd64` binary.

Installed from: `npm i -g @bazel/bazelisk --registry=https://albertocavalcante.jfrog.io/artifactory/api/npm/bazel-npm-local`

# Test Evidence

Tested on a Windows ARM4 machine.

system-about

node bazelisk.js

content of @bazel/bazelisk

Context: https://github.com/bazelbuild/bazel/issues/22164

Fixes https://github.com/bazelbuild/bazelisk/issues/572

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -1,6 +1,6 @@
-load("@rules_go//go:def.bzl", "go_binary", "go_library", "go_test")
-load("@gazelle//:def.bzl", "gazelle")
 load("@aspect_rules_js//npm:defs.bzl", "npm_package", "stamped_package_json")
+load("@gazelle//:def.bzl", "gazelle")
+load("@rules_go//go:def.bzl", "go_binary", "go_library", "go_test")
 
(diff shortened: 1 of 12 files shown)


albertocavalcante:
@meteorcloudy @fweikert can you please review?

meteorcloudy (approved):
Thanks a lot!

Khlooddeeb:
جيد

Issue #317, opened 2022-05-13
Fetch arm64 version when using USE_BAZEL_VERSION=$commit


meteorcloudy:
Currently even the arm64 version of Bazelisk fetches x86_64 version of Bazel binaries when fetching from a commit hash.

https://github.com/bazelbuild/bazelisk/blob/9102523655067867b701fc25ae9819b9acd2c24f/platforms/platforms.go#L11 needs to be updated to include `macos_arm64` and `windows_arm64`.

Although, we currently only publish arm64 Bazel binaries for macOS and Windows, not Linux ...

meteorcloudy:
@Wyverald

fweikert:
I don't have an M1 machine to test, but shouldn't line 35 handle this problem? It's also possible that DarwinFallback() cannot handle commits (just aa theory).

meteorcloudy:
Yeah, I think line 35 only handles release versions (`v.LessThan(armSupportVer)`)

keith:
See also https://github.com/bazelbuild/bazelisk/pull/303/

elfringham:
Just got bitten by this bug when our CI attempted building TensorFlow overnight for AARCH64 after an update to pick a commit hash for the bazel version.
https://ci.linaro.org/view/All/job/ldcg-python-manylinux-tensorflow-onednn-nightly/140/console

learning-to-play:
What is the status of this issue? TensorFlow supports ARM64. If `USE_BAZEL_VERSION=$commit` doesn't work on ARM64, it practically prevents TensorFlow from using this feature.

keith:
are you using the latest bazelisk release?

elfringham:
@keith We were using 1.11.0. I just tried it with 1.12.0 and get a different error now, it fails with a 404. So assuming the path now attempted to be downloaded by bazelisk is correct, then maybe this is now a bazel problem in that the artifacts are not being created.
`bazel --version
2022/08/06 09:47:13 Using unreleased version at commit f440f8ec3f63e5d663e1f9d9614f05a39422102a
2022/08/06 09:47:13 Downloading https://storage.googleapis.com/bazel-builds/artifacts/ubuntu1404_arm64/f440f8ec3f63e5d663e1f9d9614f05a39422102a/bazel...
[...]

PR #604, opened 2024-08-21
Fetch bazel binaries for commits on linux arm64

meteorcloudy:
Depends on https://github.com/bazelbuild/continuous-integration/pull/2022

Fixes https://github.com/bazelbuild/bazelisk/issues/317

Bazel commits after https://github.com/bazelbuild/bazel/commit/5aaffc3aad61b003be3996af17641afedee9a2ec will be supported for Linux arm64.

We've been publishing under both "linux" and "centos7" for x64 binaries for a long time, so it should be fine to switch the name.

### DIFF
--- a/platforms/platforms.go
+++ b/platforms/platforms.go
@@ -22,8 +22,8 @@ var supportedPlatforms = map[string]*platform{
 		HasArm64Binary: true,
 	},
 	"linux": {
-		Name:           "centos7",
-		HasArm64Binary: false,
+		Name:           "linux",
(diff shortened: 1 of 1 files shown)


fweikert (approved):

keith:
thanks! can you create a release with this in it? 🙏🏻

fweikert:
https://github.com/bazelbuild/bazelisk/releases/tag/v1.21.0 @keith

keith:
thanks!

PR #612, opened 2024-09-18
Set `BAZELISK` var to `bazelisk` executable path

fmeum:
This can be useful for scripts that want to know if they are running under Bazelisk and can also be used to run specific Bazel versions from within a Bazel run, e.g. to generate version-specific test data.


fmeum:
@meteorcloudy

fweikert (approved):

PR #624, opened 2024-10-23
feat: improve httputil error messages

jjmaestro:
* add attempt count to the timeout error message so it's clear the timeout is due to the total time accumulated during retries.

* add lastFailure to the timeout error message so we can get a clear reason for the retries (or, at the very least, the last retry). Otherwise, the error message obscures the real reason(s) for the retries and potentially misleads users into thinking that the errors could be e.g. network related (more typical of a timeout) rather than other types of errors (e.g. "tls: failed to verify certificate", etc.)

### DIFF
--- a/httputil/httputil.go
+++ b/httputil/httputil.go
@@ -107,7 +107,7 @@ func get(url, auth string) (*http.Response, error) {
 
 		nextTryAt := RetryClock.Now().Add(waitFor)
 		if nextTryAt.After(deadline) {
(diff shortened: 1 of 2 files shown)


jjmaestro:
The motivation behind this PR was my own debugging creating a Docker image where I was getting constant timeouts and I eventually figured it was due to not having `ca-certificates` installed in my Debian-based image. I debugged it by patching the error messages with this PR. 

I hope it's also useful for others!

fweikert (approved):

Issue #626, opened 2024-10-11
bazelisk sometimes fails to download Bazel binaries


werkt:
The ubuntu22 image (and probably others) have a bazelisk that precedes a number of fixes in golang with http fetches, notably identified in https://github.com/googleapis/google-cloud-go/issues/7440

```
2024/10/09 13:25:45 could not download Bazel: failed to download bazel: failed to download bazel: could not copy from https://releases.bazel.build/7.3.2/release/bazel-7.3.2-linux-x86_64 to /var/lib/buildkite-agent/.cache/bazelisk/downloads/_tmp/download1376248149: stream error: stream ID 1; INTERNAL_ERROR; received from peer
```
is the reported error, and I expect that with a newer bazelisk (current on image is 1.19, likely running against older golangs) like 1.22 with golang 1.23 basis, this situation should be improved.

meteorcloudy:
This is still happening on Bazel CI: see failed jobs in https://buildkite.com/bazel/bcr-presubmit/builds/8565#0192e691-f5e4-43fd-afda-c2b24fe85137

PR #627, opened 2024-11-05
Close response body before retry

meteorcloudy:
Hopefully fixes https://github.com/bazelbuild/bazelisk/issues/626


meteorcloudy:
I haven't found a way to reproduce the `INTERNAL_ERROR` locally.

meteorcloudy:
Actually I managed to reproduce by running `bazel info output_base` in parallel. Verifying if this change helps.

meteorcloudy:
I cannot reproduce the same error anymore after this change, looks like it worked!

fweikert (approved):

PR #635, opened 2024-11-20
Extract a method for locating Bazel

shs96c:
`bazelisk` can be thought of as two separate pieces. The first of these is a mechanism for locating a particular `bazel` version, possibly downloading it from the Web, or using a local install. The second piece is the additional functionality for flag flipping and finding out where problems have occurred within commits to the project.

This PR extracts the location of the `bazel` binary into its own method which can be used by other tooling that also needs to locate the binary.


shs96c:
Running the failing `//:go_bazelisk_test` test on `HEAD` of `master` locally also times out for me. I'm not sure the test failure is related to this PR.

fweikert (approved):

fweikert:
Thank you!

PR #649, opened 2025-01-07
Use legacy apparent names for rules_go and gazelle

fmeum:
When bazelisk is brought in as a Go dependency via gazelle's `go_deps` extension, it uses the repo mapping of the `gazelle` module to resolve loads. Since `gazelle` uses the legacy apparent names `bazel_gazelle` and `io_bazel_rules_go` for the `gazelle` and `rules_go` modules, the same is required for any Bazel module that is also a Go module (until gazelle gains the ability to fix load labels). 

This regressed in 4448ad1b1c77764b75a0e9a3d9605a7446f769d6.


fmeum:
@fweikert @meteorcloudy

meteorcloudy (approved):

PR #655, opened 2025-02-06
Hide the per os/arch binaries behind a macro.

tjgq:
This makes it easier to scrub these targets when importing into the Google monorepo.


defs.bzl
```diff
@@ -0,0 +1,20 @@
+load("@io_bazel_rules_go//go:def.bzl", "go_binary")
+
+def bazelisk_go_binaries():
+    for os in ("linux", "darwin", "windows"):
+        for arch in ("amd64", "arm64"):
+            ext = ".exe" if os == "windows" else ""
```
fweikert, line 6: Nit: move one level up

fweikert (approved):

defs.bzl
```diff
@@ -0,0 +1,19 @@
+load("@io_bazel_rules_go//go:def.bzl", "go_binary")
+
+def bazelisk_go_binaries():
```
albertocavalcante, line 3: Was it intentional to strip debug symbols from Windows binaries as well? (They're more sensitive to being flagged as suspicious malware without it).

tjgq: It wasn't intentional. I'll send a PR.

PR #658, opened 2025-02-13
Ignore signals using notify/stop instead of ignore/reset

bduffany:
See comments in https://github.com/buildbuddy-io/buildbuddy/issues/8326 - `signal.Ignore()` is "permanent" so `Reset()` has no effect.

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -701,8 +701,9 @@ func runBazel(bazel string, args []string, out io.Writer, config config.Config) 
 	// TODO(#512): We may want to treat a `bazel run` commmand differently.
 	// Since signal handlers are process-wide global state and bazelisk may be
 	// used as a library, reset the signal handlers after the process exits.
-	signal.Ignore(syscall.SIGINT, syscall.SIGQUIT, syscall.SIGTERM)
-	defer signal.Reset(syscall.SIGINT, syscall.SIGQUIT, syscall.SIGTERM)
+	sigCh := make(chan os.Signal, 1)
+	signal.Notify(sigCh, syscall.SIGINT, syscall.SIGQUIT, syscall.SIGTERM)
+	defer signal.Stop(sigCh)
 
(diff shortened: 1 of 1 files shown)


core/core.go
```diff
@@ -701,8 +701,9 @@ func runBazel(bazel string, args []string, out io.Writer, config config.Config)
 	// TODO(#512): We may want to treat a `bazel run` commmand differently.
 	// Since signal handlers are process-wide global state and bazelisk may be
 	// used as a library, reset the signal handlers after the process exits.
-	signal.Ignore(syscall.SIGINT, syscall.SIGQUIT, syscall.SIGTERM)
-	defer signal.Reset(syscall.SIGINT, syscall.SIGQUIT, syscall.SIGTERM)
+	sigCh := make(chan os.Signal, 1)
```
fmeum, line 704: I'm not sure how relevant this is in practice, but should we set the channel size to something reasonably large (perhaps just the number of signals?) based on https://cs.opensource.google/go/go/+/refs/tags/go1.24.0:src/os/signal/signal.go;l=110-113;drc=b28b263a91e4fd4af6900e045cd35c2ed1d6e3ff?

akirchhoff-modular: If the channel is full, the notification is discarded silently: https://cs.opensource.google/go/go/+/refs/tags/go1.24.0:src/os/signal/signal.go;l=244-248;drc=b28b263a91e4fd4af6900e045cd35c2ed1d6e3ff. We are attempting to ignore signals anyway, so that's fine.

bduffany: That's my understanding as well - the only reason I even set the buffer to 1 is because it triggered a nogo error, presumably to catch cases where people do want to be notified of signals.

fmeum:
@meteorcloudy @fweikert Could you review this?

fweikert (approved):

PR #667, opened 2025-03-12
Add sha256 checksums to releases

icfaust:
Helps to try and push forward #418 and helps on uxlfoundation/oneDAL#3078  I know from the issue there was a discussion of SLSA, and that Google may be using some sort of internal release process so this may be in the wrong direction. This PR attempts to patch in a file to the release called ```checksums.txt``` which creates a text file with the sha256 hashes based on github releases using free github runners.

I have not verified its operation. Any assistance would be greatly appreciated.

### DIFF
--- a/.github/workflows/checksum.yaml
+++ b/.github/workflows/checksum.yaml
@@ -1,0 +1,45 @@
+#
+# Copyright Contributors to bazelisk
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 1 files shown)


icfaust:
@fweikert Hallo, sagen Sie bitte Bescheid, ob es nützlich ist oder ob etwas geändert werden sollte. Vielen Dank für Ihre Arbeit an Bazelisk

icfaust:
@meteorcloudy I haven't received any response from the requested reviewer on the PR, am I doing something wrong? Let me know if there are any steps I can take to help this get reviewed

PR #682, opened 2025-04-15
Run buildifier to add loads

comius:
Run `buildifier --lint=fix  -r -v .`. This adds loads for the rules that were removed from Bazel 8. By Bazel 9 automatic loads will be disabled and all loads will need to be explicitly present.

Add dependency to `rules_shell` to `MODULE.bazel`.

See: https://github.com/bazelbuild/bazel/issues/25755


meteorcloudy (approved):
Thanks!

PR #692, opened 2025-05-06
Allow `--bisect` to be interrupted

fmeum:
Otherwise pressing Ctrl+C interrupts each individual Bazel command, but the bisection process itself continues.


fmeum:
@meteorcloudy

Issue #29, opened 2019-02-12
Support bash autocomplete


rickypai:
Since bazelisk downloads entire binaries, autocomplete scripts are not accessible. bazelisk could potentially also download the autocomplete scripts to a predetermined location to make it easier invoke.

jin:
+1, this will be excellent to have.

jin:
I suppose you can do this manually with instructions from https://docs.bazel.build/versions/master/completion.html#bash

```
$ bazel build //scripts:bazel-complete.bash
$ cp bazel-bin/scripts/bazel-complete.bash /etc/bash_completion.d/bazelisk

# new shell
$ bazelisk <TAB>
analyze                                                 --nodeep_execroot 
analyze-profile                                         --noexoblaze 
aquery                                                  --noexpand_configs_in_place 
--batch                                                 --noexperimental_oom_more_eagerly 
--batch_cpu_scheduling                                  --noidle_server_tasks 
--blazerc=                                              --noignore_all_rc_files 
..
```

rickypai:
yeah that's what i ended up doing... I was hoping something like a bundled bash completion script that `bazelisk` could provide, but it's not super blocking

philwo:
I would love to have this myself, but currently don't have time to implement this - if someone wants to contribute a PR, I would be very happy to review it.

sudoforge:
I'll take a stab at this when I have time over the next few days.

philwo:
@sudoforge Thanks! If you could find a solution that also works with zsh completions, I'd be extra grateful 😊 (I was a long-time bash user, but since macOS Catalina switched to zsh by default, I gave it a try and stuck with it.)

sudoforge:
I'll be looking to add completion for all shells that Bazel supports -- off the top of my head, I think that's just Bash and ZSH.

finn-ball:
+1 would be super useful to have

gibfahn:
FWIW this is done now for zsh on Homebrew installs of bazelisk: https://github.com/bazelbuild/homebrew-tap/pull/89

UebelAndre:
I've got it working on my mac using zsh but it's painfully slow. Are others experiencing the same?

UebelAndre:
Having done some comparisons between bazel and bazelisk using zsh completion on my mac, bazelisk is orders of magnitude slower.

UebelAndre:
[...]

PR #706, opened 2025-07-06
Add `bazelisk completion bash/fish` command

thii:
This command prints shell completion scripts for the active Bazel version. The completion scripts are downloaded on-demand and cached to Bazelisk CAS cache. This is to support generating completion scripts for older Bazel versions.

Closes https://github.com/bazelbuild/bazelisk/issues/29

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,6 +159,23 @@ bazelisk --bisect=~6.0.0..HEAD test //foo:bar_test
 
 Note that, Bazelisk uses prebuilt Bazel binaries at commits on the main and release branches, therefore you cannot bisect your local commits.
 
+### Shell completion
+
+Bazelisk offers a new `completion` command that allows you to generate shell
+completion scripts for Bazel commands.
+
+```shell
+bazelisk completion <shell>
+```
+
+where `<shell>` can be `bash` or `fish`.
+
+Note that the generated completion script is tied to the active Bazel version.
+
+The bazel completion scripts are taken from installer binaries. If you use a
+custom base URL, make sure the installer URLs are available alongside with
+bazel binaries.
+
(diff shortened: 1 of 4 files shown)


thii:
@fweikert @meteorcloudy Can you take a look?

meteorcloudy:
FYI, @fmeum is landing https://github.com/bazelbuild/bazel/pull/26479 in Bazel. But I guess this is still useful to support older bazel versions? @fmeum Can you take a look since you are already working on this?

fmeum:
I will take a closer look tomorrow. 

This does look like a pretty good "polyfill" for what `bazel help completion bash` will do in 8.4.0 and later.

What's the plan to handle the user switching between projects with different Bazel versions? Ideally `bazelisk` would always enable the completion script of the particular version of Bazel in use at the moment. Is that something you have thought about?

thii:
I actually haven't thought about that, just wanted to have an easy way to get the completion script of the correct bazel version (the completion script installed by system package managers like apt always installs the latest).

fmeum (approved):
Makes sense, we can always add such a script later.

Could you add the basic snippets needed to enable the completion scripts for the two supported shells (based on `bazelisk completion <shell>`)?

Otherwise this looks good (and very well tested).

jameslovesjodi4ever-dot (approved):

jameslovesjodi4ever-dot (approved):

031282:
I LL try...what is the mistake to fix it .

Στις Δευ 6 Οκτ 2025, 09:37 ο χρήστης jameslovesjodi4ever-dot <
***@***.***> έγραψε:

jameslovesjodi4ever-dot (approved):

PR #710, opened 2025-07-09
Add Configuration variable to allow using custom Authorization header

Lan-Hekary:
Hello,

I have a use case that I can't use Basic Auth to authenticate to the download server, but I can only use Bearer tokens.
so I made a new configuration variable to allow passing any custom string to Authorization header, including Bearer tokens.

I hope this helps other people.
Thanks


httputil/httputil.go
```diff
@@ -213,10 +215,16 @@ func DownloadBinary(originURL, destDir, destFile string, config config.Config) (
 		log.Printf("Downloading %s...", originURL)
 
 		var auth string = ""
-		t, err := tryFindNetrcFileCreds(u.Host)
-		if err == nil {
-			// successfully parsed netrc for given host
-			auth = t
+		if AuthHeader != "" {
```
fweikert, line 218: I think we should raise an error if both AuthHeader is set and a .netrc file exists.

Issue #731, opened 2025-10-30
Batch wrapper doesn't run if there is a space in the path and in parameters


valco1994:
I'm starting `bazelisk` with such a command line:
```
build --remote_download_outputs=all --target_pattern_file "C:\Users\User Name\..."
```
and it uses a batch script which is also located inside the 'C:\Users\User Name' folder.

I'm getting the following error:
```
'C:\Users\User' is not recognized as an internal or external command, operable program or batch file.
```

It happens due to the batch wrapper, everything works fine if I set `BAZELISK_SKIP_WRAPPER` or remove the wrapper.

I have investigated it and found that `syscall.StartProcess()` that is called through some intermediate layers from bazelisk's `core.runBazel()`, specifies both `lpApplicationName` and `lpCommandLine` for [CreateProcessW](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessw) WinAPI function. But it's an unreliable way to start a batch script, it works in some simple cases and fails in others, such as the one that I described above.

Batch scripts should be started in the following way according to [CreateProcessW documentation](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessw):
> To run a batch file, you must start the command interpreter; set lpApplicationName to cmd.exe and set lpCommandLine to the following arguments: /c plus the name of the batch file.

It would also be ok to just skip `lpApplicationName` and Windows would be able to find the interpreter automatically, but unfortunately, the Go standard library doesn't allow to omit `exec.Cmd.Path`.

So, the only proper way available for us in Go is using `cmd.exe` explicitly.

PR #732, opened 2025-10-30
Properly handle powershell and batch wrappers

### LINKED ISSUE #643: Powershell wrapper does not run
I was trying to use a bazel wrapper on Windows, so I created `tools/bazel.ps1` file in my project.

Trying to use that leads to this error:
```
2024/11/28 16:38:21 could not run Bazel: could not start Bazel: fork/exec D:\a\rules_sh\rules_sh\tests\tools\bazel.ps1: %1 is not a valid Win32 application.
```
It seems PowerShell scripts cannot be executed like native binaries; they need to be run through pwsh (PowerShell Core) or powershell (Windows PowerShell).

Using a bat wrapper for the wrapper works:
```
@echo off
pwsh -File "%~dp0tools\bazel_wrapper.ps1" %*
```

valco1994:
PowerShell scripts should be started using `powershell.exe -File ...` and batch scripts should be started using `cmd.exe /c "..."` when they are started via [CreateProcessW](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessw) (it's used under the hood to start processes on Windows in Go, see `syscall.StartProcess()`). Otherwise PowerShell scripts can't be started at all and batch scripts can't be started in some cases.

It's also important to mention, that `cmd.exe` requires quite an unusual command line, and regular command line preparation in `syscall.StartProcess()` (see `syscall.makeCmdLine()`) is not able to handle it properly and breaks it completely. So, I'm using `exec.Cmd.SysProcAttr.CmdLine` to prepare the command line myself and use it without changes in `syscall.StartProcess()` in the case of a batch wrapper.

This pull request fixes #643 and #731.

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -110,7 +110,10 @@ func RunBazeliskWithArgsFuncAndConfigAndOut(argsFunc ArgsFunc, repos *Repositori
 	// --print_env must be the first argument.
 	if len(args) > 0 && args[0] == "--print_env" {
 		// print environment variables for sub-processes
-		cmd := makeBazelCmd(bazelInstallation.Path, args, nil, config)
+		cmd, err := makeBazelCmd(bazelInstallation.Path, args, nil, config)
+		if err != nil {
+			return -1, err
+		}
 		for _, val := range cmd.Env {
 			fmt.Println(val)
 		}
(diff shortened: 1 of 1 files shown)


valco1994:
@fweikert, could you please review it?

valco1994:
@fweikert, @meteorcloudy could you, please, take a look at the proposed changes?

It's quite impolite to ignore the pull request for such a long time. After all, it fixes quite annoying bugs in your product.
If you have no resources to handle it right now, please let me know it and tell me when it can be handled approximately.

PR #734, opened 2025-11-05
Add BAZELISK_BASE_URL regression test

jwnimmer-tri:
This is a test for ea3fa7db (#722), to confirm that go and py code agree on what the environment variable does.

The recently-changed python semantics were already what go was doing, so the fix is good, just lacked a test.  FYI @chrisirhc @fweikert.


myasinmdugsiye (commented):
[emostore_i.c](https://github.com/user-attachments/files/24413454/emostore_i.c)

Issue #512, opened 2023-10-27
Bazelisk run doesn't propagate signals to the child process


jschaf:
### Reproduction

The situation to reproduce is a bit complex:

1. I have Go binary at `//erp/dev`.
2. `//erp/dev` runs other processes via bazelisk run, notably a Go-wrapper around Postgres.

The resulting process tree:

```sh
bazelisk run //erp/dev 
  # Starts a Go binary
  # The Go binary in turn calls the Go-wrapper for Postgres
  bazelisk run //db/go_postgres
    # go_postgres is a Go binary that calls the Postgres binary
    postgres
```

The problem is that if `//erp/dev` sends SIGINT directly to the `//db/go_postgres` using the Go stdlib `exec.Cmd`, bazelisk run swallows the signal.

However, if I type ctrl-c at the terminal, Postgres receives the signal. I think this is because the shell wires up the process attributes differently from Go (maybe https://stackoverflow.com/a/35435038/30900?)

### Context

As for why `bazelisk run` has this behavior:

Previously, when running `bazelisk run` under a shell, typing ctrl-c would send SIGINT twice to the child process. I think that's the same behavior I saw--processes started with Go propagate signals differently from processes started by the shell. 

https://github.com/bazelbuild/bazelisk/issues/307 noted the issue and recommended ignoring SIGINT. The resulting PR https://github.com/bazelbuild/bazelisk/pull/322 fixed the issue by ignoring SIGINT but has a few related problems

The relevant bazelisk code:
https://github.com/bazelbuild/bazelisk/blob/89e6693df1e9a9d49fe95f2814d7139d258a2f2c/core/core.go#L579

```go
func runBazel(/*omitted*/) (int, error) {
	cmd := makeBazelCmd()
	cmd.Start()

	c := make(chan os.Signal)
	signal.Notify(c, os.Interrupt, syscall.SIGTERM)
	go func() {
		s := <-c

		// Only forward SIGTERM to our child process.
		if s != os.Interrupt {
			cmd.Process.Kill()
		}
	}()
```

The two related problems:

1. The channel isn't buffered, meaning the signal delivery isn't guaranteed.
   From https://pkg.go.dev/os/signal#Notify:

   > Package signal will not block sending to c: the caller must ensure that c
   > has sufficient buffer space to keep up with the expected signal rate. For a
   > channel used for notification of just one signal value, a buffer of size 1
   > is sufficient.

   See also: https://stackoverflow.com/a/68593935/30900

[...]

Issue #556, opened 2024-02-23
bazelisk run/test hide segmentation violations


andrej-eisfeld:
### Description of the bug:

`bazel run` hides segmentation violations, e.g. it doesn't print them on the commandline. It likely just filters them out for some reason. This way bugs in the codebase might go unnoticed for some time and end up on the main development branch. It makes it also harder to figure out the root cause of crashes.

Also see the discussion on the original ticket here https://github.com/bazelbuild/bazel/issues/21470

### What's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

main.cc
```
#include <iostream>
using namespace std;
 
int main()
{
    char* str;
 
    // Stored in read only part of data segment //
    str = "GfG";
 
    // Problem:  trying to modify read only memory //
    *(str + 1) = 'n';
    return 0;
}
```

BUILD
```
cc_binary(
     name = "main",
     srcs = ["main.cc"],
)
```

When we "bazel run" the cc_binary here's what we get:
![image](https://github.com/bazelbuild/bazel/assets/55919036/f6e62f83-fa29-409b-bd07-0801d785495b)
![image](https://github.com/bazelbuild/bazel/assets/55919036/dbd4f1ed-a630-455e-85b6-99ae0ab7d813)

Running the binary from bazel-bin instead we clearly see the application segfaults (return code 139)
![image](https://github.com/bazelbuild/bazel/assets/55919036/6dec67d2-85cc-4d16-97bd-3ca041233ffa)

PR #757, opened 2026-01-10
Report a proper exit code when Bazel is terminated by signal and use `exec` on Unix  in the case of `bazelisk run`

valco1994:
It fixes issues  #512 and #556

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -180,7 +180,7 @@ func RunBazeliskWithArgsFuncAndConfigAndOut(argsFunc ArgsFunc, repos *Repositori
 		return 0, nil
 	}
 
-	exitCode, err := runBazel(bazelInstallation.Path, args, out, config)
+	exitCode, err := runBazel(bazelInstallation.Path, args, out, config, isRunCommand(args))
 	if err != nil {
 		return -1, fmt.Errorf("could not run Bazel: %v", err)
 	}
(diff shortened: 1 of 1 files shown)


fweikert (approved):

aaron-skydio:
@valco1994 why only exec when we find a `run` in the arg list (presumably indicating you're doing `bazel run ...`)?  Why not exec always like I did in https://github.com/bazelbuild/bazelisk/pull/566 (except when bazelisk itself has more work to do after)?

This seems especially confusing if I'm reading correctly and this will exec if you do like `bazel --config run build` or something, which is totally valid I think (equivalent to `bazel --config=run build`)

valco1994:
@aaron-skydio, I missed your pull request, and found it only when my pull request was already merged. So, I didn't consider making `exec` as often as possible, as was proposed by you in #566.

Considering the motivation, I wanted to fix issues reported by users - #512 and #556 - and in both cases, problematic scenarios were related to `bazelisk run`. Also, bazelisk is supposed to be a transparent wrapper around Bazel, and Bazel uses `exec` in the case of `bazel run`. It was impossible to rely on this behaviour before, when bazelisk was used, and it's possible now after my changes.

I have nothing against using `exec` more broadly, but probably it's worth thinking about scenarios which it can improve. Because we can't use `exec` always and must support both approaches anyway.

aaron-skydio:
Sorry for taking forever here, but I've rebased #566 to master again on top of this - the implementation there is still the same as it was originally, i.e. using `exec` whenever bazelisk doesn't intend to run anything after invoking bazel, and not relying on parsing the command line arguments to decide whether to `exec`

PR #791, opened 2026-04-27
bazelisk.py supports bazeliskVersion

jwnimmer-tri:
Also add missing regression test of the Go implementation.  Amends #763.


jwnimmer-tri:
Ping @fweikert for review, please?

PR #792, opened 2026-04-28
bazelisk.py uses os.exec

jwnimmer-tri:
Same as #566, but for the Python implementation.

### DIFF
--- a/bazelisk.py
+++ b/bazelisk.py
@@ -494,8 +494,17 @@ def make_bazel_cmd(bazel_path, argv):
 def execute_bazel(bazel_path, argv):
     cmd = make_bazel_cmd(bazel_path, argv)
 
-    # We cannot use close_fds on Windows, so disable it there.
-    p = subprocess.Popen([cmd["exec"]] + cmd["args"], close_fds=os.name != "nt", env=cmd["env"])
+    # If we can replace our process with the Bazel executable, that's better
+    # than a subprocess. (Replacing is not available on Windows.)
+    if os.name != "nt":
+        os.execve(
+            cmd["exec"],
+            [cmd["exec"]] + cmd["args"],
+            cmd["env"],
+        )
+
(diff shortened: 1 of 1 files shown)


jwnimmer-tri:
Ping @fweikert for review, please?

bazelisk.py
```diff
@@ -494,8 +494,17 @@ def make_bazel_cmd(bazel_path, argv):
 def execute_bazel(bazel_path, argv):
     cmd = make_bazel_cmd(bazel_path, argv)
 
-    # We cannot use close_fds on Windows, so disable it there.
-    p = subprocess.Popen([cmd["exec"]] + cmd["args"], close_fds=os.name != "nt", env=cmd["env"])
+    # If we can replace our process with the Bazel executable, that's better
```
fweikert: Please flush stdout and stderr before

bazelisk.py
```diff
@@ -494,8 +494,17 @@ def make_bazel_cmd(bazel_path, argv):
...
+    # than a subprocess. (Replacing is not available on Windows.)
+    if os.name != "nt":
+        os.execve(
```
fweikert: Please handle potential exceptions

fweikert (commented):
I'm a bit worried that this change may be disruptive / unexpected to people who use bazelisk.py only as library.

bazelisk.py
jwnimmer-tri: Done

bazelisk.py
jwnimmer-tri: Done

jwnimmer-tri:
> I'm a bit worried that this change may be disruptive / unexpected to people who use bazelisk.py only as library.

Interesting!  I never considered anyone doing that.  Anyway, I've updated the code now so that only command-line invocation will be affected.

PR #801, opened 2026-07-02
bazelisk has been packaged in Fedora

blinxen:
I have [packaged `bazelisk` for Fedora](https://src.fedoraproject.org/rpms/bazelisk) and it should be available to everyone in a couple of days on Fedora 44, Fedora rawhide already has it.


fweikert:
Thank you! We still need the CLA to be signed, though.

Issue #523, opened 2023-12-21
.bazelversion not being read in parent directory outside of WORKSPACE 


cpsauer:
Hi wonderful bazelisk folks,

First and foremost, thanks for a stupendous launcher. Bazelisk is really valuable and I really appreciate all the work you've done here.

I'm writing in to report a bug--or at least an instance where the docs are inconsistent with bazelisk's behavior.

The readme says:
> Otherwise, if a .bazelversion file exists in the current directory or recursively any parent directory, it will read the file and use the version specified in it.

but Bazelisk seems to stop that search at a workspace boundary rather than *any* parent directory.

I think the documented behavior is probably the more desirable one. Projects often have a nested test workspace, and it'd be nice to inherit from the parent workspace. At least that was true of https://github.com/Neargye/magic_enum, where this was causing confusion. And I'd imagine any hardcoding of WORKSPACE files might need updating with the bzlmod changes, where a WORKSPACE file is no longer required. (But there might be other good reasons for the other behavior! In that case, the docs could use a change.) 

Thanks so much for all you do,
Chris
(ex-Googler)

avdv:
I just stumbled over the same issue, but wanted to add that the file is also not read from the *current directory* if there is no WORKSPACE file.

In our case we want to check the users Bazel version for a start script and print out a warning if it's outside the expected version range, but on CI we want to use a fixed version so a `.bazelversion` file is created. 

This happens before the start script is called, and causes bazelisk to download the latest Bazel version, which will currently result in a version warning, but the start script goes on to create a Bazel project at which point bazelisk switches to the configured version (downloading that) and uses that.

mering:
Related: #502

visglz:
I can confirm that this is still an issue with bazelisk 1.25.0.

jkanywhere:
Still a problem with bazelisk 1.29.0.

keith:
https://github.com/bazelbuild/bazelisk/pull/805

Issue #52, opened 2019-04-22
Version in .bazelversion should be respected without WORKSPACE file


bmuschko:
## Issue

Executing the pre-built bazelisk binary doesn't seem to respect the value in a `.bazelversion` file when executed from the directory containing the file. The binary downloads the latest version of Bazel instead. Setting `USE_BAZEL_VERSION` picks up the correct version from the environment variable.

I believe the issue is that the determination of the `.bazelversion` file relies on the existence of a `WORKSPACE` file. For some of the "infrastructure" commands e.g. `version`, `help`, `info` this shouldn't be required.

## How to reproduce?

1. [Download bazelisk binary](https://github.com/philwo/bazelisk/releases/tag/v0.0.4) (in my case `bazelisk-darwin-amd64`)
2. Add the binary to PATH and add execution permission

```
cp bazelisk /Users/bmuschko/dev/tools/bazelisk
chmod +x /Users/bmuschko/dev/tools/bazelisk
export PATH=/Users/bmuschko/dev/tools:$PATH
```

3. Create a new directory and navigate to it

```
mkdir my-project
cd my-project
```

4. Add `.bazelversion` with version 0.21.0

```
echo 0.21.0 > .bazelversion
```

5. Run `bazelisk version`

```
Build label: 0.24.1
Build target: bazel-out/darwin-opt/bin/src/main/java/com/google/devtools/build/lib/bazel/BazelServer_deploy.jar
Build time: Tue Apr 2 16:32:47 2019 (1554222767)
Build timestamp: 1554222767
Build timestamp as int: 1554222767
```

philwo:
I think we can change the logic to check for `.bazelversion` files from the current working directory up to the workspace root, but I'm not really sure what the use case is.

Except for `bazel help` or `bazel version` you can't do anything without a WORKSPACE file - not even `bazel info`:

```
$ bazel info
ERROR: The 'info' command is only supported from within a workspace (below a directory having a WORKSPACE file).
See documentation at https://docs.bazel.build/versions/master/build-ref.html#workspace
```

Could you explain a bit more why you'd need this?

philwo:
@bmuschko Friendly ping - I'm curious about your use case for this. :)

bmuschko:
@philwo When I opened the issue, I didn't realize that a WORKSPACE is pretty much required no matter what as I was very new to Bazel. IMHO it's not a use case. ;-)

ian-h-chamberlain:
[...]

PR #804, opened 2026-07-15
Respect .bazelversion file for batch mode

keith:
Some commands like 'bazel verison' can be run outside of a bazel
workspace. In this case the .bazelversion file in the PWD was ignored
because there is no MODULE.bazel file. Now in this fallback case we
attempt to use PWD instead. This is potentially useful for
pre-downloading bazel in container images without having your entire
repo present.


jkanywhere:
https://github.com/bazelbuild/bazelisk/issues/523 and https://github.com/bazelbuild/bazelisk/issues/52 report the same problem.

keith:
yea I'm going to do a more targeted fix for #523 since I didn't see that before submitting this

PR #805, opened 2026-07-15
Look for config files anywhere up the tree

### LINKED ISSUE #52: Version in .bazelversion should be respected without WORKSPACE file

### LINKED ISSUE #523: .bazelversion not being read in parent directory outside of WORKSPACE

keith:
Previously the README said that bazelisk would discover the
`.bazelversion` file "in the current directory or recursively any parent
directory", but this wasn't the case. It would only find it in the root
of the project. This doesn't work for cases where users have this layout
for testing:

```
repo/MODULE.bazel
repo/.bazelversion
repo/e2e/MODULE.bazel
```

In this case it would only look beside `repo/e2e/MODULE.bazel`, not find
the `repo/.bazelversion` and fallback to other logic. Folks often fix
this by symlinking the root `.bazelversion` to the subdirectory.

This changes the behavior to do what was stated before. The new behavior
applies for the `.bazeliskrc` too.

Fixes https://github.com/bazelbuild/bazelisk/issues/52
Fixes https://github.com/bazelbuild/bazelisk/issues/523


keith:
I chose not to touch `repo/tools/bazel` discovery here since that felt a bit weird to do recursively up, but maybe that leads to more confusion?

jkanywhere (approved):

core/core.go
```diff
@@ -413,26 +413,23 @@ func GetBazelVersion(config config.Config) (string, error) {
...
-			scanner.Scan()
-			bazelVersion := scanner.Text()
-			if err := scanner.Err(); err != nil {
-				return "", fmt.Errorf("could not read version from file %s: %v", bazelVersion, err)
-			}
+		scanner := bufio.NewScanner(f)
+		scanner.Scan()
+		bazelVersion := scanner.Text()
+		if err := scanner.Err(); err != nil {
+			return "", fmt.Errorf("could not read version from file %s: %v", bazelVersion, err)
```
fweikert: Please use bazelVersionPath instead (it's a pre-existing bug, but would be nice to fix here).

fweikert (approved):

keith:
updated!

fweikert (approved):

PR #808, opened 2026-08-08
Add bash completion conffile to debian packages

jwnimmer-tri:


jwnimmer-tri:
This is an improvement upon #563, so I'll ping @fweikert for review, please?

deb/BUILD
```diff
@@ -17,6 +19,31 @@ genrule(
...
 
+run_binary(
+    name = "bazelisk_completion_bash",
+    args = [
+        "completion",
+        "bash",
+    ],
+    env = {
+        "HOME": "/tmp",
+        "USE_BAZEL_VERSION": "9.2.0",
```
fweikert: Maybe replace with latest to avoid this becoming stale

fweikert (changes requested):

deb/BUILD
jwnimmer-tri: The upstream (bazel) completion script doesn't change that often, so I don't think staleness is a _huge_ concern.  (Previously, users would install this conffile by hand, so it would necessarily become stale since it was a manual process to refresh it.)

I do think this should be pinned to an exact version, so that `bazelisk` releases are reproducible when rebuilt from the same tag.  That seems more important than possible staleness.

I'm open to suggestions on how we could (via docs and/or automation and/or CI scripting) encourage this pin to be keep somewhat current.

PR #814, opened 2026-09-17
Support patches of release candidates

### LINKED ISSUE #813: support patches of release candidates
I needed to patch Bazel 9.3.0rc1 in order to try it out.  So naturally, I named the patched release candidate something like 9.3.0rc1-patch1.

I then quickly discovered that Bazelisk has no concept of a patched release candidate: https://github.com/bazelbuild/bazelisk/blob/master/versions/versions.go#L22-L23

via a diagnostic:
```
2026/09/17 08:35:43 could not download Bazel: could not resolve the version '9.3.0rc1-patch1' to an actual version number: invalid version '9.3.0rc1-patch1'
```

dws:
This permits naming a patched release candidate in a reasonable way, such as "9.3.0rc1-patch1".  This is analogous to how we already support naming patches of releases, such as "9.3.0-patch1".

Fixes #813.


versions/versions.go
```diff
@@ -68,6 +69,9 @@ func Parse(fork, version string) (*Info, error) {
 	} else if candidatePattern.MatchString(version) {
 		vi.IsLTS = true
 		vi.MustBeCandidate = true
+	} else if candidatePatchPattern.MatchString(version) {
```
fweikert: Maybe merge the if clause with the previous one since the body is the same.

fweikert (changes requested):

versions/versions.go
dws: Done!

fweikert (approved):

PR #815, opened 2026-09-19
Document `BAZELISK_WRAPPER_PATH`

marlamb:
It was introduced more than two years ago in #567 . Documenting it
explicitly makes it easier to find and highlights that it is a part of
the public interface.


fweikert (approved):
Note: the PR title contains an incorrect env var name

